PR #4, opened 2017-10-16
better log trimming

mrose17:
and more accurate notice on uphold registration


eyeshade/controllers/publishers.js
```diff
@@ -243,6 +243,140 @@ v2.getBalance = {
...
+        contributions: {
+          amount: amount,
+          currency: currency,
+          altcurrency: altcurrency,
+          probi: probi.truncated().toString()
+        }
+      }
+
+      entries = await settlements.find({ publisher: publisher }, { sort: { timestamp: -1 }, limit: 1 })
+      console.log(JSON.stringify(entries, null, 2))
```
evq: debug?

eyeshade/controllers/publishers.js
```diff
@@ -243,6 +243,140 @@ v2.getBalance = {
...
+      entry = entries && entries[0]
+      if (entry) {
+        probi = new BigNumber(entry.probi.toString())
+        result.lastSettlement = {
+          amount: runtime.currency.alt2fiat(altcurrency, probi, currency) || 0,
```
evq: should we also include the fiat value at the time of conversion in the settlement objects? maybe just for cards denominated in fiat?

eyeshade/controllers/publishers.js
```diff
@@ -243,6 +243,140 @@ v2.getBalance = {
...
+        {
+          $group:
+          {
+            _id: '$publisher',
+            probi: { $sum: '$probi' }
+          }
+        }
+      ])
+      if (summary.length > 0) probi = probi.minus(new BigNumber(summary[0].probi.toString()))
+      if (probi.lessThan(0)) probi = 0
```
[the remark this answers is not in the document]
mrose17: +1

bat-utils/lib/runtime-wallet.js
```diff
@@ -124,27 +124,34 @@ Wallet.prototype.unsignedTx = async function (info, amount, currency, balance) {
   return f.bind(this)(info, amount, currency, balance)
 }
 
+Wallet.prototype.submitTx = async function (info, txn, signature) {
```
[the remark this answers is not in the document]
mrose17: yes, i was trying to get the functions to line up with the cases below

eyeshade/controllers/publishers.js
mrose17: interesting idea. let me look into that.

mrose17:
@evq - back to you!

eyeshade/controllers/publishers.js
```diff
@@ -803,12 +941,15 @@ module.exports.initialize = async (debug, runtime) => {
      // v2 and later
         altcurrency: '',
         probi: bson.Decimal128.POSITIVE_ZERO,
+        currency: '',
+        amount: bson.Decimal128.POSITIVE_ZERO,
```
evq: we could probably get away with a regular number here instead of Decimal128 since this value is approx / for display. either is fine though

eyeshade/workers/wallet.js
```diff
@@ -116,6 +116,8 @@ exports.initialize = async (debug, runtime) => {
 
 /*
```
evq: I think we should avoid committing WIP code in commented out form

eyeshade/controllers/publishers.js
mrose17: sorry, "bump"?

perhaps. i just prefer these things to be consistent....

eyeshade/workers/wallet.js
mrose17: it's not WIP. it's what i use when debugging things. i'll take it out.

eyeshade/controllers/publishers.js
evq: bump was on the earlier comment to log to sentry if `probi.lessThan(0)` - not sure why it didn't thread correctly. no worries, sgtm

mrose17: that may not be an exception. with the bitcoin proof-of-concept, there was a one-time "thing" where i ran settlement twice. it's OK for probi to be zero in cases like that.

mrose17:
@evq - anything else i should look at? thanks!

evq:
@mrose17 there's still a duplicate line - https://github.com/brave-intl/bat-ledger/blob/ea00aae9c91965bccb8e1d7990ac3539fd3d0724/eyeshade/controllers/publishers.js#L117

eyeshade/controllers/publishers.js
evq: Sorry I'm confused. Is the probi for publishers who went through double settlement negative or did it get zeroed? (I'm assuming it's okay to be zero means it's okay to zero here.) I still think it's a good idea to capture the case where it is negative. Obviously we'll get some events due to the double settlement, but over time assuming continued contributions we should expect all to go positive -> so an uptick in this event would be a red flag

mrose17: i've added a call to `runtime.captureException`

evq:
lgtm, staging integration tests are go

PR #47, opened 2017-11-14
switch to constant time comparison for simple token auth

### LINKED ISSUE #44: non-constant time comparison used for simple token auth
might be hard to pull off due to latencies, but I think there's a potential side channel attack
@diracdeltas fyi

evq:
fixes #44

### DIFF
--- a/bat-utils/lib/extras-hapi.js
+++ b/bat-utils/lib/extras-hapi.js
@@ -1,3 +1,4 @@
+const crypto = require('crypto')
 const ProxyAgent = require('proxy-agent')
 const SDebug = require('sdebug')
 const underscore = require('underscore')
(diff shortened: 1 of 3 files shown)


bat-utils/lib/extras-hapi.js
```diff
@@ -32,6 +33,31 @@ exports.domainCompare = (a, b) => {
   }
 }
 
+const constantTimeEquals = (a, b) => {
+  let mismatch = a.length !== b.length
+  if (mismatch) {
+    b = a
```
evq: would it be better to do `b = mismatch ? a : b`? there's probably a slight diff between runtime with equal length strings and nonequal due to the extra assignment as it stands.

idk if an assignment to self would be optimized out or not

mrose17: well, that's the problem with optimizers. sometimes they do the wrong thing for the right reasons.

evq:
thanks for the review @mrose17 !

will wait for @diracdeltas to take a look, then we can go ahead with merge

bat-utils/lib/extras-hapi.js
diracdeltas: @BrendanEich may be the best to answer that question :). however, I don't think it's necessary to prevent a timing attack against length of input, so just `crypto.timingSafeEqual(Buffer.from(a), Buffer.from(b))` would be sufficient. if you really wanted to hide the token length, i think it's better to just HMAC the inputs and compare those (since they would then be equal length).

evq: makes sense. `timingSafeEqual` throws an error on unequal length inputs so will stick with the current code to ensure they are equal

PR #48, opened 2017-11-14
replace bittrex websockets with binance websockets

mrose17:


package.json
```diff
@@ -107,9 +107,9 @@
   "devDependencies": {
     "ava": "^0.22.0",
     "babel-eslint": "^8.0.0",
-    "bitgo": "4.12.0",
+    "bitgo": "^4.6.0",
```
evq: `4.13.0` is broken if you don't have a ssh keypair which is registered with github where you are doing `npm install`, thus the pin at `4.12.0`

PR #54, opened 2017-11-15
add the bat-helper server

mrose17:
to run service, configure `TOKEN_LIST`, `HELPER_GITHUB_CLIENT_ID` and `HELPER_GITHUB_CLIENT_SECRET`

to use service in `bat-ledger` or `bat-eyeshade`, configure `HELPER_URL` and `HELPER_TOKEN`

Fixes https://github.com/brave-intl/bat-ledger/issues/55


config.js
```diff
@@ -1,106 +1,132 @@
...
+
+  balance: {
+    portno: 3003,
+
+    f: () => {
+      if (process.env.LEDGER_URL) {
+        module.exports.ledger = { url : process.env.LEDGER_URL  || 'http://127.0.0.1:3001' }
+      }
+
+      helper()
```
evq: we still need uphold config if we are supporting the card based apis for older builds

docker-compose.yml
```diff
@@ -87,3 +87,33 @@ services:
...
+      - "SERVICE=helper"
+    depends_on:
+      - mongo
+      - redis
+    networks:
+      - ledger
+  helper-web:
+    container_name: helper-web
+    ports:
+      - "3002:3002"
```
evq: s/3002/3004

eyeshade/controllers/rates.js
```diff
@@ -0,0 +1,43 @@
...
+   GET /v1/rates
+ */
+
+v1.read = { handler: (runtime) => {
+  return async (request, reply) => {
+    const result = { altrates: runtime.currency.altrates, fxrates: runtime.currency.fxrates, rates: runtime.currency.rates }
+    let validity
+
+    underscore.keys(result).forEach((key) => { if (underscore.keys(result[key]).length === 0) delete result[key] })
+    validity = Joi.validate(result, runtime.currency.schemas.rates)
```
evq: is there a reason to explicitly validate here vs using the response validation below?

docker-compose.yml
mrose17: good catch!

eyeshade/controllers/rates.js
mrose17: another good catch! there's a hook that i can use.

config.js
mrose17: good point.

mrose17:
@evq - back to you... thanks for the thoughtful review!

PR #56, opened 2017-11-15
HOTFIX: when a channel is bulk-loaded, be sure to send a...

mrose17:
...publisher-report to the back-end, so the ledger is informed.


mrose17:
good point. with the exception of the `unique` property, the fields are really their for completeness. even so, i'll fix that..

PR #60, opened 2017-11-16
HF: fix publisher status caching algorithm

mrose17:
eyeshade periodically goes to the "source of truth" (publishers) to get contact information and visible status. unfortunately, when doing the comparison and deciding whether to update the local datastore and let ledger know, eyeshade wasn't looking at the visible status.


eyeshade/workers/reports.js
```diff
@@ -132,27 +132,27 @@ const hourly2 = async (debug, runtime) => {
...
           state = {
             $currentDate: { timestamp: { $type: 'timestamp' } },
             $set: { info: info }
           }
           if (visibleP) state.$set.visible = visible
-          if (method) state.$set.method = method
+          params = underscore.pick(record, [ 'info', 'visible' ])
+          if (method) {
+            state.$set.method = method
+            params.method = record.method
```
evq: can we just pick this above? to keep things in one place

mrose17:
OBE by cherrypick.

PR #63, opened 2017-11-20
Staging

mrose17:
and cleanup runtime-currency


evq:
going to merge via push

PR #64, opened 2017-11-21
Initial check-in

mrose17:
For some reason, the helper branch didn’t end up in staging, so here is a clean merge.


mrose17:
no worries, thanks!

PR #74, opened 2017-11-29
incl publicKey in grant redeem req, tweak test, add grant srv to compose

evq:
the integration between the ledger service and the grant redemption service has now been tested end to end and is working :tada:


mrose17:
_hooray!_

PR #75, opened 2017-11-29
if a paymentId is provided...

mrose17:
…then make sure the wallet exists, otherwise it’s 404-for-you!


mrose17:
This is for @bsclifton

mrose17:
time to make @bsclifton a writer.

evq:
++

PR #80, opened 2017-12-04
invalidate balance on successful grant claim

evq:


ledger/controllers/grants.js
```diff
@@ -208,6 +208,18 @@ v1.write = { handler: (runtime) => {
...
+      await braveHapi.wreck.delete(runtime.config.balance.url + '/v2/wallet/' + paymentId + '/balance',
+        {
+          headers: {
+            authorization: 'Bearer ' + runtime.config.balance.access_token,
+            'content-type': 'application/json'
+          },
+          useProxyP: true
+        })
+    }
+
```
mrose17: i suggest putting this in a `try ... catch` which sentry-logs the failure...

evq: good idea :+1:

PR #81, opened 2017-12-05
bail out of redeem if there are no grants assigned to a wallet

evq:


bat-utils/lib/runtime-wallet.js
```diff
@@ -152,6 +152,8 @@ Wallet.prototype.redeem = async function (info, txn, signature) {
 
   if (!this.runtime.config.redeemer) return
 
+  if (!info.grants) return
+
```
mrose17: this works for me!

PR #82, opened 2017-12-05
add wallet runtime back for balance server

evq:


balance/server.js
```diff
@@ -15,6 +15,5 @@ const options = {
 config.database = false
 config.login = false
 config.queue = false
-config.wallet = false
```
mrose17: understood.

PR #86, opened 2017-12-07
When calculating the exchange rate for a single probi...

mrose17:
…do not apply scaling. Apply it later to avoid zero-dot-zero-zero

### DIFF
--- a/bat-utils/lib/runtime-currency.js
+++ b/bat-utils/lib/runtime-currency.js
@@ -305,9 +305,11 @@ const monitor3 = (config, runtime) => {
 
 const maintenance = async (config, runtime) => {
   const now = underscore.now()
-  let fxrates, results, tickers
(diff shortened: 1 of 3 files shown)


bat-utils/lib/runtime-currency.js
```diff
@@ -302,9 +305,11 @@ const monitor3 = (config, runtime) => {
 
 const maintenance = async (config, runtime) => {
   const now = underscore.now()
-  let fxrates, results, tickers
+  let fxrates, rates, results, tickers
 
   if (config.helper) {
+    rates = JSON.parse(JSON.stringify(singleton.rates))
```
evq: ?

mrose17: that is the "easy" way to deep-clone a non-cyclic javascript object. i can require underscore-deepclone if you prefer.

PR #91, opened 2017-12-14
Add /v1/address/{alt currency}/{address} and /v1/wallet/{paymentId} endpoints

mrose17:
To help with wallet forensics

### DIFF
--- a/eyeshade/controllers/address.js
+++ b/eyeshade/controllers/address.js
@@ -9,7 +9,7 @@ const braveJoi = utils.extras.joi
 const v1 = {}
 
 /*
-   GET /v1/address/{paymentId}
+   GET /v1/address/{altcurrency}/{address}
  */
 
 v1.get =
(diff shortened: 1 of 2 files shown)


eyeshade/controllers/address.js
```diff
@@ -0,0 +1,60 @@
...
+const underscore = require('underscore')
+
+const utils = require('bat-utils')
+const braveHapi = utils.extras.hapi
+const braveJoi = utils.extras.joi
+
+const v1 = {}
+
+/*
+   GET /v1/address/{paymentId}
```
evq: this doesn't match

eyeshade/controllers/wallet.js
```diff
@@ -0,0 +1,86 @@
...
+      paymentId: Joi.string().guid().required().description('identity of the wallet')
+    }
+  },
+
+  response: {
+    schema: Joi.object().keys({}).unknown(true)
+  }
+}
+
+v1.get =
```
evq: copy pasta?

mrose17: thanks for catching this. worse than that: it's the 2016 mbp keyboard...

PR #94, opened 2017-12-21
add voting cohort support, control and grant cohorts

evq:


eyeshade/workers/reports.js
```diff
@@ -1240,7 +1240,8 @@ exports.workers = {
             publisher: slice.publisher,
             votes: slice.counts,
             created: new Date(parseInt(slice._id.toHexString().substring(0, 8), 16) * 1000).getTime(),
-            modified: (slice.timestamp.high_ * 1000) + (slice.timestamp.low_ / bson.Timestamp.TWO_PWR_32_DBL_)
+            modified: (slice.timestamp.high_ * 1000) + (slice.timestamp.low_ / bson.Timestamp.TWO_PWR_32_DBL_),
+            cohort: slice.cohort || 'control'
```
mrose17: this is why i don't like the `standard` rule for where to put commas between properties. if we could put them at the beginning of the line, then the only diff here would be the line `, cohort: ...`

evq: something, something trailing commas :)

ledger/controllers/surveyor.js
```diff
@@ -493,61 +493,43 @@ module.exports.create = create
...
-      const params = {
-        surveyorId: entry.surveyorId,
-        avail: entry.surveyors.length,
-        viewingId: viewing.viewingId,
-        needed: viewing.count
-      }
-      if (viewing.surveyorIds.length >= params.needed) return debug('fixup not needed', params)
+      count = desiredCount - cSurveyor.cohorts[cohort].length
+      while (count > 0) {
+        vSurveyor = await create(debug, runtime, 'voting', { cohort: cohort }, cSurveyor.surveyorId)
```
evq: @mrose17 here is where we associate a surveyor with a cohort

eyeshade/workers/reports.js
mrose17: _heh._ however, there is a serious point here: i'd rather see one green line and not a red line followed by two green ones, where the only difference between the red and first green line is the addition of a comma. it is visually distracting otherwise.

ledger/controllers/surveyor.js
mrose17: thanks

eyeshade/workers/reports.js
evq: :P yeah i'm agreeing that a single line change would be better, I just wish it was via a trailing comma after the final element instead of leading comma before each element

mrose17: understood, i think the issue is that seeing a comma at EOL suggests there should be something on the next line. seeing a comma and then text disambiguates that...

PR #96, opened 2017-12-21
Owner model updates

mrose17:
- add `POST /v2/owners` for both websites and oauth publishers
- add `GET /v1/owners/{owner}/verify/{publisher}` and `PUT /v1/owners/{owner}/verify/{publisher}` for website publisher verification
- deprecate `POST /v1/publishers`
- add `GET /v1/publishers/identity?url=…` to eyeshade to replace `GET/v2/publishers/identity?url=…` on ledger
- put oauth name/email into corresponding publisher entries
- handle per-owner reports

### DIFF
--- a/eyeshade/controllers/owners.js
+++ b/eyeshade/controllers/owners.js
@@ -558,8 +558,8 @@ v1.getToken = {
 v1.putToken = {
   handler: (runtime) => {
     return async (request, reply) => {
-      return putToken(request, reply, runtime, null, request.params.owner, request.params.publisher,
-                      request.payload.verificationId, request.query.show_verification_status)
+      return putToken(request, reply, runtime, request.params.owner, request.params.publisher, request.payload.verificationId,
+                      request.query.show_verification_status)
     }
   },
 
--- a/eyeshade/controllers/owners.js
+++ b/eyeshade/controllers/owners.js
@@ -13,9 +13,26 @@ const utils = require('bat-utils')
 const braveHapi = utils.extras.hapi
 const braveJoi = utils.extras.joi
 
+const verifier = require('./publishers.js')
+const getToken = verifier.getToken
+const putToken = verifier.putToken
+
 const v1 = {}
+const v2 = {}
 
 let altcurrency
+
+const ownerString = (owner, info) => {
+  const name = info && (info.name || info.ownerName)
+  const email = info && (info.email || info.ownerEmail)
+  let result = name
+
+  if (result && email) result += ' <' + email + '>'
+  if (result) result += ' '
+  result += owner
+
+  return result
+}
 
 /*
    POST /v1/owners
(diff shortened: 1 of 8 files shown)


evq:
@ayumi fyi

eyeshade/controllers/owners.js
```diff
@@ -44,14 +44,15 @@ v1.bulk = {
...
+      const channels = []
+      let info = {}
 
-      if (authorizer.ownerEmail || authorizer.ownerName) {
-        extras = underscore.omit(authorizer, [ 'owner' ])
-        providers.forEach((provider) => { underscore.extend(provider, extras) })
-      }
+      if (authorizer.ownerEmail || authorizer.ownerName) info = { email: authorizer.ownerEmail, name: authorizer.ownerName }
+      providers.forEach((provider) => {
+        channels.push(underscore.extend({ channeId: provider.publisher, visible: provider.show_verification_status }, info))
```
evq: `channeId` - typo?

eyeshade/controllers/owners.js
```diff
@@ -44,14 +44,15 @@ v1.bulk = {
...
       const providers = request.payload.providers || []
-      let extras
+      const channels = []
+      let info = {}
 
-      if (authorizer.ownerEmail || authorizer.ownerName) {
-        extras = underscore.omit(authorizer, [ 'owner' ])
-        providers.forEach((provider) => { underscore.extend(provider, extras) })
-      }
+      if (authorizer.ownerEmail || authorizer.ownerName) info = { email: authorizer.ownerEmail, name: authorizer.ownerName }
```
evq: should this be `&&`?

eyeshade/controllers/owners.js
```diff
@@ -154,46 +159,100 @@ const bulk = async (request, reply, runtime, owner, info, providers) => {
...
 
   reply({})
 }
 
+/*
+   DELETE /v1/owners/{owner}/{publisher}
+       [ used by publishers ]
+ */
+
+v1.deletePublisher = {
```
evq: nit: maybe `unlinkPublisher` here to match description?

eyeshade/controllers/owners.js
```diff
@@ -154,46 +159,100 @@ const bulk = async (request, reply, runtime, owner, info, providers) => {
...
+  description: 'Unlinks a publisher from an owner',
+  tags: [ 'api' ],
+
+  validate: {
+    params: {
+      owner: braveJoi.string().owner().required().description('the owner identity'),
+      publisher: braveJoi.string().publisher().required().description('the publisher identity')
+    },
+    query: {
+      access_token: Joi.string().guid().optional()
```
evq: is this used?

mrose17: it allows a person using swagger to test the endpoint!

eyeshade/controllers/owners.js
mrose17: i don't think it really matters, but since i'm going to edit the file anyway... why not?

eyeshade/controllers/owners.js
mrose17: not a typo. the folks at tilde requested that we use `channelId` for the `/v2/...` endpoint. since this is the `/v1/...` endpoint and they both use the same worker routine, we convert it.

eyeshade/controllers/owners.js
mrose17: it should be `||` if either is set, use the either/both values. if either, one will be `undefined` which worksin this case.

eyeshade/controllers/owners.js
evq: ah I see. that was not obvious to me since this is part of the query and not the authorization header as I would expect.

maybe we do it as a header validation instead? e.g.

```
validate: {
      headers: Joi.object({
        'authorization': Joi.string().required(),
      }).unknown()
}
```

I believe that would make it show up as an available field on swagger ala:
https://github.com/glennjones/hapi-swagger/issues/435

evq:
looks good pending transcripts, thanks!

mrose17:
not ready yet

PR #101, opened 2018-01-03
Simplex cc integration

mrose17:
if `runtime.config.simplex` is defined via the usual configuration mechanism, then information necessary for a `POST` to purchase BTC  for the browser's BAT wallet is returned on `GET /v2/wallet/{paymentId}`

### DIFF
--- a/bat-utils/lib/runtime-wallet.js
+++ b/bat-utils/lib/runtime-wallet.js
@@ -581,11 +581,6 @@ Wallet.providers.simplex = {
     })
     if (Buffer.isBuffer(result)) try { result = JSON.parse(result) } catch (ex) { result = result.toString() }
 
-// courtesy of https://stackoverflow.com/questions/2076299/how-to-close-current-tab-in-a-browser-window#18776480
-// javascript:window.open('','_self').close();
-// courtesy of http://www.yournewdesigner.com/css-experiments/javascript-window-close-firefox.html
-// javascript:window.open('','_parent','').close();
-
     return ({
       quotes: { simplex: quote },
       buyForm: {
--- a/bat-utils/lib/runtime-wallet.js
+++ b/bat-utils/lib/runtime-wallet.js
@@ -526,7 +526,7 @@ Wallet.providers.simplex = {
           'content-type': 'application/json'
         },
         payload: JSON.stringify({
-          end_user_id: info.paymentId,
+          end_user_id: quote ? quote.user_id : uuid.v4().toLowerCase(),
           digital_currency: params.currency,
           fiat_currency: currency,
           requested_currency: currency,
@@ -587,7 +587,7 @@ Wallet.providers.simplex = {
 // javascript:window.open('','_parent','').close();
 
     return ({
-      extend: { simplex: quote },
+      quotes: { simplex: quote },
       buyForm: {
         method: 'POST',
         action: this.runtime.config.simplex.url + '/payments/new',
--- a/bat-utils/lib/runtime-wallet.js
+++ b/bat-utils/lib/runtime-wallet.js
@@ -4,6 +4,7 @@ const UpholdSDK = require('@uphold/uphold-sdk-javascript')
 const bitcoinjs = require('bitcoinjs-lib')
 const crypto = require('crypto')
 const underscore = require('underscore')
+const uuid = require('uuid')
 const { verify } = require('http-request-signature')
 
 const braveHapi = require('./extras-hapi')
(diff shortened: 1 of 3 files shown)


bat-utils/lib/runtime-wallet.js
```diff
@@ -475,4 +496,118 @@ Wallet.providers.mock = {
...
+// courtesy of https://stackoverflow.com/questions/2076299/how-to-close-current-tab-in-a-browser-window#18776480
+// javascript:window.open('','_self').close();
+// courtesy of http://www.yournewdesigner.com/css-experiments/javascript-window-close-firefox.html
+// javascript:window.open('','_parent','').close();
+
+    return ({
+      extend: { simplex: quote },
+      buyForm: {
+        method: 'POST',
+        action: this.runtime.config.simplex.url + '/payments/new',
```
evq: updating with the results on our out of band discussion:
I would prefer that we did not make this url server provided but it was discussed and we have decided to proceed as implemented, we will enforce that the url provided is https in the browser

mrose17: i would be happy to add a signature over the URL provided by a third-party that the browser could verify in a subsequent revision.

bat-utils/lib/runtime-wallet.js
```diff
@@ -475,4 +496,118 @@ Wallet.providers.mock = {
...
+        authorization: 'ApiKey ' + this.runtime.config.simplex.api_key,
+        'content-type': 'application/json'
+      },
+      payload: JSON.stringify({
+        account_details: {
+          app_provider_id: params.partner,
+          app_version_id: params.version,
+          app_end_user_id: quote.user_id,
+          signup_login: {
+            ip: '213.162.55.5',
```
evq: what is the significance of this ip?

bat-utils/lib/runtime-wallet.js
```diff
@@ -475,4 +496,118 @@ Wallet.providers.mock = {
...
+              currency: params.currency,
+              address: info.addresses.BTC
+            }
+          }
+        }
+      })
+    })
+    if (Buffer.isBuffer(result)) try { result = JSON.parse(result) } catch (ex) { result = result.toString() }
+
+// courtesy of https://stackoverflow.com/questions/2076299/how-to-close-current-tab-in-a-browser-window#18776480
```
evq: can we remove these comments? they don't seem related

bat-utils/lib/runtime-wallet.js
```diff
@@ -475,4 +496,118 @@ Wallet.providers.mock = {
...
+    quote = info.simplex
+    if ((!quote) || (quote.expires <= now)) {
+      result = await braveHapi.wreck.post(this.runtime.config.simplex.url + '/wallet/merchant/v2/quote', {
+        headers: {
+          accept: 'application/json',
+          authorization: 'ApiKey ' + this.runtime.config.simplex.api_key,
+          'content-type': 'application/json'
+        },
+        payload: JSON.stringify({
+          end_user_id: info.paymentId,
```
evq: is this the user wallet's paymentId? do we really need to send that to simplex? can we deterministicly generate an id only used with simplex?

ledger/controllers/wallet.js
```diff
@@ -110,9 +111,23 @@ const read = function (runtime, apiVersion) {
               $set: { unsignedTx: result.unsignedTx }
             }
           }
-          await wallets.update({ paymentId: paymentId }, state, { upsert: true })
         }
       }
+
```
evq: I would prefer we set `state=null` here rather then clear after use above

bat-utils/lib/runtime-wallet.js
```diff
@@ -475,4 +496,118 @@ Wallet.providers.mock = {
...
+    })
+    if (Buffer.isBuffer(result)) try { result = JSON.parse(result) } catch (ex) { result = result.toString() }
+
+// courtesy of https://stackoverflow.com/questions/2076299/how-to-close-current-tab-in-a-browser-window#18776480
+// javascript:window.open('','_self').close();
+// courtesy of http://www.yournewdesigner.com/css-experiments/javascript-window-close-firefox.html
+// javascript:window.open('','_parent','').close();
+
+    return ({
+      extend: { simplex: quote },
```
evq: I would prefer a more descriptive name than extend. How about `walletQuote`?

bat-utils/lib/runtime-wallet.js
mrose17: the significance is that it is a non-existent IP address, cf., https://tools.ietf.org/html/rfc1918

we aren't going to supply the browser's IP address to the provider.

bat-utils/lib/runtime-wallet.js
mrose17: i plan to after the browser-side implementation is done. these links are the result of "considerable" research in how to use javascript to close the current tab. i thought it would be as simple as `_self.close()`, but nothing is that simple, sadly.

bat-utils/lib/runtime-wallet.js
mrose17: we can certainly add another stable UUID. i will do that.

bat-utils/lib/runtime-wallet.js
mrose17: ok.

ledger/controllers/wallet.js
mrose17: why? the logical place to put it is after it is used, so if it isn't set again, we know we don't have to save it?

bat-utils/lib/runtime-wallet.js
evq: there are other mediums more suited for sharing and storing this research

ledger/controllers/wallet.js
evq: my thinking is that clearing it right before use is defensive since it makes it less likely that we add an additional branch that doesn't clear it

bat-utils/lib/runtime-wallet.js
evq: sure, the browser will do that itself when it follows the link to go to the provider page :p 

I was wondering in the context of simplex but based on your answer it's the expected ip of the user signing up and can be wrong / spoofed without issue.

PR #103, opened 2018-01-04
HOTFIX: do not exit if no trades reported during a cycle

mrose17:
…. just log it and proceed.

### DIFF
--- a/bat-utils/lib/runtime-currency.js
+++ b/bat-utils/lib/runtime-currency.js
@@ -340,6 +340,7 @@ const maintenance = async (config, runtime) => {
   if (flatlineP) {
     debug('maintenance', { message: 'no trades reported' })
     runtime.captureException(new Error('maintenance reports flatline'))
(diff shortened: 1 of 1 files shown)


bat-utils/lib/runtime-currency.js
```diff
@@ -340,7 +340,6 @@ const maintenance = async (config, runtime) => {
   if (flatlineP) {
     debug('maintenance', { message: 'no trades reported' })
     runtime.captureException(new Error('maintenance reports flatline'))
-    process.exit(0)
```
evq: we were only hitting this branch on select services that were misconfigured to not use the helper service to retrieve exchange rate information. should we still drop this?

how do you feel about failing out if helper is not configured when env = prod?

mrose17: the `process.exit` is called only when helper is not configured (look at the top of the function!) . i've made the other change you suggest...

PR #104, opened 2018-01-05
Re-enable green check-marks

mrose17:
Complete the separation of `verified` (green check-mark) and `visible` (opted-in to marketing use).

Current:

    % curl -s -X GET --header 'Accept: application/json' 'https://ledger.mercury.basicattentiontoken.org/v3/publisher/identity?publisher=nickvonpentz.com' | json_pp
    {
    "QLD" : "",
    "properties" : {
    "timestamp" : "6493179644546646017"
    },
    "SLD" : "nickvonpentz.com",
    "publisher" : "nickvonpentz.com",
    "RLD" : ""
    }

New:

    % curl -s -X GET --header 'Accept: application/json' 'http://127.0.0.1:3001/v3/publisher/identity?publisher=nickvonpentz.com' | json_pp
    {
    "publisher" : "nickvonpentz.com",
    "SLD" : "nickvonpentz.com",
    "properties" : {
    "verified" : true,
    "timestamp" : "6501003756404998145"
    },
    "QLD" : "",
    "RLD" : ""
    }

### DIFF
--- a/ledger/controllers/publisher.js
+++ b/ledger/controllers/publisher.js
@@ -640,6 +640,13 @@ module.exports.initialize = async (debug, runtime) => {
       unique: [ { rulesetId: 1 } ],
       others: [ { type: 1 }, { version: 1 }, { timestamp: 1 } ]
     },
+/* verified publishers
+   - verified should always be "true"
+   - visible indicates whether the publisher opted-in to inclusion in marketing materials
+
+   originally this was the 'publishers' table, but was renamed to 'publishersX' to temporarily address a publisher privacy
+   issue. however, it was accidentally commented out, which resulted in vanilla servers not getting the indices...
+ */
     {
       category: runtime.database.get('publishersX', debug),
       name: 'publishersX',
--- a/ledger/controllers/publisher.js
+++ b/ledger/controllers/publisher.js
@@ -480,7 +480,7 @@ v3.identity =
         if ((timestamp) && ((!result.properties.timestamp) || (timestamp > result.properties.timestamp))) {
           result.properties.timestamp = timestamp
         }
-        if (entry.visible) result.properties.verified = entry.verified
+        if (entry.verified) result.properties.verified = entry.verified
       }
 
       reply(result)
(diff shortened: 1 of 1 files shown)


ledger/controllers/publisher.js
```diff
@@ -640,16 +640,14 @@ module.exports.initialize = async (debug, runtime) => {
       unique: [ { rulesetId: 1 } ],
       others: [ { type: 1 }, { version: 1 }, { timestamp: 1 } ]
     },
-/* deprecated
     {
       category: runtime.database.get('publishers', debug),
```
evq: should this be `get('publishersX'...` ? we have two collections in use as far as I can tell - `publishersV2` and `publishersX`. additionally there is one legacy collection `publishers`.

the code here looks like it's creating an index `publishersX` on the legacy collection `publishers`. is that right? seems confusing to reuse the collection name as an index name on a different collection.

ledger/controllers/publisher.js
```diff
@@ -640,16 +640,14 @@ module.exports.initialize = async (debug, runtime) => {
       unique: [ { rulesetId: 1 } ],
       others: [ { type: 1 }, { version: 1 }, { timestamp: 1 } ]
     },
-/* deprecated
```
evq: maybe it would be good to add some comments explaining the difference between the various collections here. it's hard to understand the intent / what's still in use without digging through code as is.

a general style comment - I think we should generally avoid commenting out blocks of code like this (and above) in the future. it's easier to grep the code to tell what's still in use and if we ever need to restore them we can simply pull them out of the git history. for cases where it makes more sense to leave the code in place but commented we should leave a descriptive comment. `deprecated` isn't enough, I would have guessed that this block was gone and never coming back and that we were just too lazy to remove it based on that comment. but here we are restoring it - not obvious.

mrose17: agreed.

ledger/controllers/publisher.js
mrose17: sorry, my fault: `publishers` became `publishersX` and the edit below should have just just included the rename rather that commenting it out. i botched that.

evq:
did you re-test since the last change? how about adding some comments?

mrose17:
yes, i tested it to ensure it created `pubishersX` and associated indices.

i will add "some comments"

evq:
thanks! by "some comments" I meant the ones I suggested earlier that you said you agreed with

PR #106, opened 2018-01-09
do not invoke /me/cards if user.status === 'pending'

mrose17:
as per the discussion with uphold

### DIFF
--- a/bat-utils/lib/runtime-wallet.js
+++ b/bat-utils/lib/runtime-wallet.js
@@ -348,7 +348,7 @@ Wallet.providers.uphold = {
     }
   },
   status: async function (info) {
-    let card, cards, currency, currencies, result, uphold, user
+    let authorized, card, cards, currency, currencies, result, uphold, user
 
     try {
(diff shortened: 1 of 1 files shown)


bat-utils/lib/runtime-wallet.js
```diff
@@ -359,7 +359,8 @@ Wallet.providers.uphold = {
       uphold.storage.setItem('uphold.access_token', info.parameters.access_token)
 
       user = await uphold.api('/me')
-      cards = await uphold.api('/me/cards')
+      authorized = user.status !== 'pending'
```
evq: this is less restrictive than the check it's replacing. I assume this is correct since you talked with uphold, however I can't check myself since their documentation doesn't explain possible values of this field as far as I can tell

mrose17: yes, they said that the test was to make sure the status was *not* `"pending"` ... hence this change.  and yes, the SDK doc could be a little clearer.

PR #110, opened 2018-01-10
Add authorizer.visible to set default for owner visibility

mrose17:
And the default for the associated publisher(s).


mrose17:
OBE

PR #111, opened 2018-01-11
If we don't have settlement address/currency for a publisher

mrose17:
Send a better notice to the #publishers-bot and do not include the entry in the settlement file.


mrose17:
cf., https://bravesoftware.slack.com/archives/C2QCSFWSW/p1515704015000732 et. seq., for examples.

PR #112, opened 2018-01-12
Add /v2/reports/publishers/statements/{settlementId}'

mrose17:
cf.,
https://bravesoftware.slack.com/files/U0C84CCAC/F8SJCE13Q/publishers-statements-20180111-161055-413.csv

### DIFF
--- a/eyeshade/controllers/reports.js
+++ b/eyeshade/controllers/reports.js
@@ -12,6 +12,7 @@ const braveJoi = utils.extras.joi
 
 const v1 = {}
 const v2 = {}
+const v3 = {}
 
 let altcurrency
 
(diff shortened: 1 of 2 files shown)


eyeshade/controllers/reports.js
```diff
@@ -481,6 +523,7 @@ module.exports.routes = [
   braveHapi.routes.async().path('/v1/reports/publishers/settlements').config(v1.publishers.settlements),
   braveHapi.routes.async().path('/v1/reports/publisher/{publisher}/statements').config(v1.publisher.statements),
   braveHapi.routes.async().path('/v1/reports/publishers/statements/{hash}').config(v1.publishers.statements),
+  braveHapi.routes.async().path('/v2/reports/publishers/statements/{settlementId}').config(v2.publisher.statements),
```
evq: v2.publisher.statements doesn't seem to fit the existing naming pattern

mrose17: for /v1, we use a BTC `{hash}` for /v2 we use a `{settlementId}` ...

evq: sure but it's plural `v1.publishers.statements` vs singular `v2.publisher.statements`

PR #113, opened 2018-01-16
Ensure that the currency is USD

mrose17:
And that the minimum USD is enforced…

Test: when 1BAT=0.477USD, a request for 5BAT is translated into a request for 5USD (first, ceil’d up to 2, then bumped to 5)

### DIFF
--- a/bat-utils/lib/runtime-wallet.js
+++ b/bat-utils/lib/runtime-wallet.js
@@ -505,7 +505,8 @@ Wallet.providers.simplex = {
       partner: 'brave',
       version: '1'
     }
-    let expires, min, quote, rate, result
+    min = this.runtime.config.simplex && (this.runtime.config.simplex['MIN_' + fiat.toUpperCase()] || 5)
+    let expires, quote, rate, result
 
     if (!this.runtime.config.simplex) return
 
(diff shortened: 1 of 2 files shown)


bat-utils/lib/runtime-wallet.js
```diff
@@ -514,8 +514,15 @@ Wallet.providers.simplex = {
...
       amount = Math.ceil(this.currency.alt2fiat(currency, amount, fiat, true))
-      currency = fiat
+    } else if (currency !== fiat) {
+      rate = underscore.pick(this.currency.fxrates[currency])
+      if (!rate) return
+
+      amount = Math.ceil(amount / rate)
     }
+    currency = fiat
+    min = this.runtime.config.simplex['MIN_' + fiat.toUpperCase()] || 5
```
NejcZdovc: where do we set this?

NejcZdovc: this can be const

mrose17: by `this` do you mean `this.runtime...`? the answer is that it is set during server initialzation. look at `config.js`

i have made `min` a constant... however, my coding style is to put `const` at the beginning of a block, so you'll notice the assignment is a bit more noisy.

NejcZdovc: I mean that I don't see MIN_USD in config.js

mrose17: right. the default is 5. if we want to override that, we can add something to config.js

NejcZdovc: ok so this is just for future use, got it

PR #117, opened 2018-01-17
Update GET wallet to know about the current defaults

mrose17:
1. Both the ledger and balance servers will return a `properties` object
2. To aid debugging, the balance server now uses swagger and `/v1/login`
3. When the current defaults are changed, the ledger will tell the
balance server to update the current defaults; similarly, when the
balance server starts, it will ask the ledger for the current defaults
4. The “current defaults” are the `payload` associated with the current
`persona registrar`.

### DIFF
--- a/balance/controllers/address.js
+++ b/balance/controllers/address.js
@@ -14,6 +14,8 @@ BigNumber.config({ EXPONENTIAL_AT: 28 })
 const debug = new SDebug('balance')
 const v2 = {}
 
+let properties = {}
+
 /*
    GET /v2/card/BAT/{cardId}/balance
  */
@@ -51,6 +53,7 @@ v2.batBalance =
       unconfirmed: balanceProbi.minus(spendableProbi).dividedBy(runtime.currency.alt2scale(altcurrency)).toFixed(4),
       rates: runtime.currency.rates[altcurrency]
     }
+    if (underscore.keys(properties).length) underscore.extend(balances, { properties: properties })
 
     reply(balances)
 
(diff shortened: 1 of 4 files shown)


ledger/controllers/wallet.js
```diff
@@ -47,6 +47,9 @@ const read = function (runtime, apiVersion) {
 
     if (apiVersion === 2) {
       result = underscore.extend(result, { addresses: wallet.addresses })
+      if (runtime.registrars.persona) {
+        result = underscore.extend(result, { parameters: runtime.registrars.persona.payload || {} })
```
evq: shouldn't this be `properties` to match the balance response schema?

mrose17: good question. let me check!

mrose17: it should be `parameters` ... fixed! thanks for catching that.

balance/controllers/address.js
```diff
@@ -14,6 +14,8 @@ BigNumber.config({ EXPONENTIAL_AT: 28 })
 const debug = new SDebug('balance')
 const v2 = {}
 
+let parameters = {}
```
evq: this won't work if we're running multiple workers - which we are in prod

evq: (we'll only update one worker's copy)

mrose17: should be fixed!

balance/controllers/address.js
```diff
@@ -176,3 +177,15 @@ module.exports.routes = [
   braveHapi.routes.async().path('/v2/wallet/{paymentId}/balance').config(v2.walletBalance),
   braveHapi.routes.async().delete().path('/v2/wallet/{paymentId}/balance').config(v2.invalidateWalletBalance)
 ]
+
+module.exports.initialize = async (debug, runtime) => {
```
evq: I don't think we need this anymore, otherwise lgtm

PR #128, opened 2018-01-24
fix passing user agent if doesn't exist

evq:


evq:
hmm. it seems better to me to pass the correct type in the first place versus cleaning it up after

mrose17:
it seems better to me to fix it once in one place and as generally as possible.

evq:
I agree, we should switch to a statically typed language :)

mrose17:
as long as it is also a functional programming language, why not? (-;

but that train left the station two years ago...

evq:
in liu of fixing this in the most general way, my thoughts here are that I would much prefer an error when calling something with the wrong type versus having the called code try to fix my types. having the code fixup the types by deleting keys is rather unintuitive.

evq:
let me know if you disagree and still pref deleting keys, I'll change it if so

mrose17:
i disagree with your approach. i greatly prefer the approach i outlined...

bat-utils/lib/extras-hapi.js
```diff
@@ -156,6 +156,10 @@ const WreckProxy = (server, opts) => {
   if (!opts.headers) opts.headers = {}
   if (!opts.headers['user-agent'] && !opts.headers['User-Agent']) opts.headers['user-agent'] = wreckUA
 
+  underscore.keys(opts.headers).forEach((header) => {
+    if (typeof opt.headers[header] !== 'string') delete opt.headers[header]
```
evq: typo - will repush

PR #129, opened 2018-01-28
Generate statistics for grafana directlly

mrose17:
cf., https://github.com/grafana/grafana/blob/master/docs/sources/plugins/developing/datasources.md and https://github.com/grafana/simple-json-datasource


evq:
I don't think this is the right approach

mrose17:
@evq - perhaps later today you can help me understand your thinking on this... thanks!

PR #131, opened 2018-01-30
Transition update!

mrose17:
1. Delete old owners at the end of the process (previous code tried to do this _in situ_ and missed quite a few).
2. Be sure to update the `authority` field for owners and publishers to the new ownerId.

### DIFF
--- a/eyeshade/controllers/owners.js
+++ b/eyeshade/controllers/owners.js
@@ -188,7 +188,7 @@ v3.bulk = {
           entry = await publishers.findOne({ publisher: channelId })
           if (!entry) continue
 
-          if ((entry.owner) && (cleanup.indexOf(entry.owner))) cleanup.push(entry.owner)
+          if ((entry.owner) && (cleanup.indexOf(entry.owner) == -1)) cleanup.push(entry.owner)
           await publishers.update({ publisher: channelId }, {
             $set: { owner: owner.owner_identifier, authority: owner.owner_identifier }
           }, { upsert: true })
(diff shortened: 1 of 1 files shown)


eyeshade/controllers/owners.js
```diff
@@ -187,8 +188,10 @@ v3.bulk = {
           entry = await publishers.findOne({ publisher: channelId })
           if (!entry) continue
 
-          await publishers.update({ publisher: channelId }, { $set: { owner: owner.owner_identifier } },
-                                  { upsert: true })
+          if ((entry.owner) && (cleanup.indexOf(entry.owner))) cleanup.push(entry.owner)
```
ayumi: `cleanup` starts as `[]` so isn't `cleanup.indexOf(entry.owner)` going to return false and prevent pushing of owners to cleanup every time?

eyeshade/controllers/owners.js
```diff
@@ -187,8 +188,10 @@ v3.bulk = {
...
+          await publishers.update({ publisher: channelId }, {
+            $set: { owner: owner.owner_identifier, authority: owner.owner_identifier }
```
ayumi: why do we set two properties both to the same value?

mrose17: in the future, there may be a different entity that authorizes an owner...

eyeshade/controllers/owners.js
ayumi: * updated my comment above

mrose17: that's a typo. fixed

Issue #97, opened 2018-01-02
move source-of-truth for publisher rules to eyeshade


mrose17:
1. Deprecate these calls in the ledger:

    GET    /v2/publisher/identity       
    GET    /v2/publisher/identity/verified

2. Move these calls from the ledger to the eyeshade server:

    POST   /v2/publisher/ruleset
    PATCH  /v2/publisher/rulesets
    PUT    /v2/publisher/ruleset/{publisher}
    DELETE /v2/publisher/ruleset/{publisher}
    GET    /v2/reports/publisher/rulesets

3. Add messages from eyeshade to ledger that provide the functionality in the five calls above.

mmclau14:
@mrose17 do we want to `move` or copy and deprecate the old so that people using this endpoint can have time to move over? or will the 404 be enough to inform the individual at that time

mmclau14:
1. taken care of in commit: `36f77d1a`

mrose17:
OBE.

Issue #65, opened 2017-11-21
remove v1 legacy api support (bitcoin)


evq:

mrose17:
+1

mrose17:
we had a bitcoin proof-of-concept a few years back. good for learning things. when we moved to BAT, we changed a lot of API endpoints from /v1 to /v2, since we couldn't get the same semantics. what this issue refers to is simply removing the BTC-only v1 API calls.  if you look at the admin page for the ledger, you'll see a number of calls tarting with /{apiV}/... that do either v1 (BTC proof-of-concept) or v2 (BAT Mercury).

mmclau14:
alright, so just hard code all of the `/{apiV}/` to `/v2/` easy enough

mrose17:
i think step 1 is to remove the /v1's that are accessible under /{apiV}

i think step 2 is to ask @evq whether we should get rid of /{apiV} or leave it. we might leave it, because we may have /v3 in the future. she will know.

mmclau14:
cool cool. can do

evq:
When doing some other cleanup I noticed that there are still v1 wallet endpoints, @maikelmclauflin can you remove them?

PR #142, opened 2018-02-21
Collector update

mrose17:
This is a refactoring update. The changes are:

1. database and queue initialization is centralized in a `{SERVICE}/common.js` file for each service. before it was spread across various files and sometimes duplicate.

2. multiple database instances are now supported, because collector needs a read-write database for `pseries` and read-only access to the other collections

3. the report file creation is now centralized in the database runtime

4. multiple queue instances are also supported, though this is not currently used

5. things which are common to a particular service's controller or workers, can also be found in `{SERVICE}/controllers/common.js` `{SERVICE}/workers/common.js`

6. `deprecated` tags are added to the REST entrypoints that are no longer used by the publishers serivce

7. a little bit of reordering was done for the scripts in `package.json`


mrose17:
once this PR is merged, then these issues can be addressed: https://github.com/brave-intl/bat-ledger/issues/65, https://github.com/brave-intl/bat-ledger/issues/97, and https://github.com/brave-intl/bat-ledger/issues/130

mrose17:
OBE.

PR #144, opened 2018-02-22
add report option to retrieve excluded votes for testng

evq:


eyeshade/controllers/reports.js
```diff
@@ -505,7 +505,8 @@ v1.surveyors.contributions = {
   validate: {
     query: {
       format: Joi.string().valid('json', 'csv').optional().default('csv').description('the format of the report'),
-      summary: Joi.boolean().optional().default(true).description('summarize report')
+      summary: Joi.boolean().optional().default(true).description('summarize report'),
+      excluded: Joi.boolean().optional().default(false).description('only include excluded in report')
```
mrose17: just one small nit: "include only excluded entries in report"

the existing wording is "poor" english (with respect to the placement of "only")...

PR #146, opened 2018-02-27
Report verification failures

mrose17:
Report verification failures, just like we report verification success.

cc: @lgebhardt

### DIFF
--- a/eyeshade/controllers/publishers.js
+++ b/eyeshade/controllers/publishers.js
@@ -961,11 +961,11 @@ const verified = async (request, reply, runtime, entry, verified, backgroundP, r
     $set: { verified: entry.verified, reason: reason.substr(0, 64) }
   }
   await tokens.update(indices, state, { upsert: true })
-  if (!verified) return
 
   reason = reason || (verified ? 'ok' : 'unknown')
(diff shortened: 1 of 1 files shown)


eyeshade/controllers/publishers.js
```diff
@@ -961,11 +961,11 @@ const verified = async (request, reply, runtime, entry, verified, backgroundP, r
     $set: { verified: entry.verified, reason: reason.substr(0, 64) }
   }
   await tokens.update(indices, state, { upsert: true })
-  if (!verified) return
 
   reason = reason || (verified ? 'ok' : 'unknown')
   payload = underscore.extend(underscore.pick(entry, [ 'verificationId', 'token', 'verified' ]), { status: reason })
   await publish(debug, runtime, 'patch', entry.owner, entry.publisher, '/verifications', payload)
```
ayumi: can we be 300% sure `verified` is either `true` or `false`?

Publishers checks if this value is truthy:
https://github.com/brave-intl/publishers/blob/staging/app/controllers/api/channels_controller.rb#L43

mrose17: verified is ether true or false, no exceptions.

PR #149, opened 2018-03-02
When doing sanity checks

mrose17:
when looking at a verification token entry, if the publisher entry hasn’t been created, then the token entry MUST have `verified=false`

@ayumi - this should prevent future 404s from occurring; however, we need to deal with the existing 404s. the eyeshade entries with them can not be recovered, sadly... let's discuss on #publishers ... thanks!


mrose17:
sorry! i'm working on solutions to these issues.

diracdeltas:
@ayumi why did you approve the PR in that case?

PR #150, opened 2018-03-03
must also delete verifications tokens when deleting a publisher

mrose17:
Otherwise, a new owner can not be used to verify the publisher


eyeshade/controllers/owners.js
```diff
@@ -352,6 +352,8 @@ v1.unlinkPublisher = {
       }
       await publishers.update({ publisher: publisher }, state, { upsert: true })
 
+      await tokens.remove({ publisher: publisher }, { justOne: false })
```
aekeus: Is the `tokens` variable in scope here?

mrose17: commited the *wrong* file... sigh. fixed.

mrose17: sorry and many thanks!

PR #158, opened 2018-03-13
Consult each index individually...

mrose17:
monk:queries [worker@1104 id="1"] grants.indexes: {"_id_":[["_id",1]],"altcurrency_1":[["altcurrency",1]],"grantId_1":[["grantId",1]],"promotionId_1":[["promotionId",1]],"timestamp_1":[["timestamp",1]],"paymentId_":[["paymentId",""]],"probi_1":[["probi",1]]} +3ms
        worker [worker@1104 id="1"] grants indices already created +0ms
        monk:queries [worker@1104 id="1"] voting.indexes: {"_id_":[["_id",1]],"surveyorId_1_publisher_1_cohort_1":[["surveyorId",1],["publisher",1],["cohort",1]],"timestamp_1":[["timestamp",1]],"counts_1":[["counts",1]],"publisher_1_altcurrency_1_exclude_1_probi_1":[["publisher",1],["altcurrency",1],["exclude",1],["probi",1]]} +1ms
        worker [worker@1104 id="1"] voting indices being updated +0ms
        monk:queries [worker@1104 id="1"] voting.createIndex: "exclude_1" +17ms
        monk:queries [worker@1104 id="1"] voting.createIndex: "hash_1" +27ms
        monk:queries [worker@1104 id="1"] voting.createIndex: "altcurrency_1_probi_1" +33ms
        monk:queries [worker@1104 id="1"] voting.createIndex: "altcurrency_1_exclude_1_probi_1" +22ms
        monk:queries [worker@1104 id="1"] voting.createIndex: "owner_1_altcurrency_1_exclude_1_probi_1" +25ms


diracdeltas:
could you explain what this is for?

bat-utils/lib/runtime-database.js
```diff
@@ -128,27 +128,51 @@ Database.prototype.get = function (collection, debug) {
 Database.prototype.checkIndices = async function (debug, entries) {
   entries.forEach(async (entry) => {
     const category = entry.category
-    let doneP, indices
+    let doneP, indices, status
 
-    try { indices = await category.indexes() } catch (ex) { indices = [] }
-    doneP = underscore.keys(indices).indexOf(entry.property + '_1') !== -1
+    const form = (index) => {
```
diracdeltas: why is this called `form`? consider using a more descriptive variable name to make it easier to understand what is going on

bat-utils/lib/runtime-database.js
```diff
@@ -128,27 +128,51 @@ Database.prototype.get = function (collection, debug) {
...
+      let result = ''
 
-    debug(entry.name + ' indices ' + (doneP ? 'already' : 'being') + ' created')
+      underscore.keys(index).forEach((key) => { result += '_' + key + '_' + index[key] })
+      return result.substr(1)
+    }
+
+    const gather = (list) => {
```
diracdeltas: why do `form` and `gather` need to be defined inside every entries.forEach call? can you hoist them outside the scope into utility functions?

mrose17: good catch! they don't. i will move them up one scope.

bat-utils/lib/runtime-database.js
mrose17: i like `form`. the reason i don't use something like `key` or `keys` or `keyed` or something more descriptive is because those are already used in the prototype and i am trying to avoid confusion.

mrose17:
when there is more staffing/cycles/etc., unit tests are appropriate. for now, under the agreement reached earlier with devops, showing the code running in the description (forgot to included it earlier, it's there now), is considered sufficient!

bat-utils/lib/runtime-database.js
mrose17: i'm happy to change `form` to something else, if you have a suggestion!

mrose17:
_could you explain what this is for?_ the previous version of the code looked to see if the primary index was defined; if so, it didn't check the other indices. however, mongo reported some slow queries that could be fixed by more indices, so these were added in an earlier PR. however, that requires that the code here check each index to see if it exists, and if not, to create it.

flamsmark:
@mrose17: I don't understand what this PR is about, or why I'm a qualified reviewer. Is there an issue documenting the problem that this fixes? I'm certainly not deeply familiar with the ledger codebase.

mrose17:
@flamsmark - you're right. there's another review i should put you on. i will, don't worry (-;

mrose17:
@diracdeltas - the current workload i have doesn't give me the luxury to do the unit test at this time. there are only so many live grenades i can juggle before i drop one. the good news is that we have been hiring to address this need, and we have someone coming in who is going to be able to help quite a bit. for now, however; i am continuing under the compromise reached earlier with devops.

diracdeltas:
@mrose17 i refactored your code to be unit-testable and added a unit test in https://github.com/brave-intl/bat-ledger/pull/158/commits/db6788b1e3653f00a37dd7e844920d227dfa47f6. please see TODO comments in it. hopefully this helps.

mrose17:
@diracdeltas - thanks. a couple of things. i had to make a minor edit to the code so it would run (using `this` inside an anonymous function -- this this case, `gather` -- never does what is desired. i also added some whitespace to reflect my coding style.

in terms of the tests: none of the existing tests actually run. doing `npm run test-unit` resulted in an error as well, as one of the tests was a no-op.

all fixed now.

bat-utils/lib/runtime-database.js
```diff
@@ -125,30 +125,61 @@ Database.prototype.get = function (collection, debug) {
...
+  underscore.keys(index).forEach((key) => { result += '_' + key + '_' + index[key] })
+  return result.substr(1)
+}
+
+// TODO: annotate this function and give it a more descriptive name
+Database.prototype.gather = (entry) => {
+  const gather = (list) => {
+    const result = []
+
+    if (list) list.forEach((index) => { result.push(Database.prototype.form(index)) })
```
diracdeltas: good catch on my `this` binding failure. as @BrendanEich points out however it's best to not use arrow functions in prototype for this reason

test/bat-utils.unit.test.js
```diff
@@ -23,6 +24,42 @@ test('bat-utils : isSimpleTokenValid', async t => {
...
+    unique: [ { surveyorId: 1, publisher: 1, cohort: 1 } ],
+    others: [ { counts: 1 }, { timestamp: 1 },
+              { exclude: 1 }, { hash: 1 }, { counts: 1 },
+              { altcurrency: 1, probi: 1 },
+              { altcurrency: 1, exclude: 1, probi: 1 },
+              { owner: 1, altcurrency: 1, exclude: 1, probi: 1 },
+              { publisher: 1, altcurrency: 1, exclude: 1, probi: 1 } ]
+  }
+
+  t.is(JSON.stringify(db.gather(entry)), JSON.stringify([
```
diracdeltas: you can use `t.deepEqual` to compare objects without stringifying

mrose17: thanks!

bat-utils/lib/runtime-database.js
```diff
@@ -125,30 +125,59 @@ Database.prototype.get = function (collection, debug) {
...
+
+    if (list) list.forEach(function (index) { result.push(this.form(index)) }.bind(this))
+
+    return result
+  }
+
+  return gather(entry.unique).concat(gather(entry.others), gather(entry.raw))
+}
+
 Database.prototype.checkIndices = async function (debug, entries) {
```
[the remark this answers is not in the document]
mrose17: i spent a few years "doing rails". they have this "thing" about having database "migration" scripts. (i don't have the heart to tell them that the word they want to use is "transition", not "migration" -- birds migrate back-and-forth, database schemas should not.) 

in any event, one thing that i learned during that era that it is much better to have the application make sure that the schema, indices, etc., are current when the application starts. that is what `checkIndices` does. the organization is a little more clear in the [extractor branch](https://github.com/brave-intl/bat-ledger/tree/extractor) which does a better job of DRY.

PR #159, opened 2018-03-14
Add more publishers support

mrose17:
Use `/api/tokens/` endpoint on publishers to receive all tokens
awaiting verification.


nvonpentz:
@mmclau14

evq:
since we're moving site verification over to publishers per https://github.com/brave-intl/publishers/issues/815 I think we can close this out

PR #163, opened 2018-04-03
remember/utilize preferred currency from publishers

mrose17:
And set owner in settlement records.


mrose17:
we aren't going to use hyphenated names. if you'd rather call it defaultCurrency rather than preferredCurrency, that's fine with me.

PR #165, opened 2018-04-03
add helper function for setting mongo schema for collection

evq:
also adds a schema for the `restricted` collection as an example

### DIFF
--- a/bat-utils/lib/runtime-database.js
+++ b/bat-utils/lib/runtime-database.js
@@ -185,4 +185,22 @@ Database.prototype.checkIndices = async function (debug, entries) {
   })
 }
 
+/**
+ * Sets the schema for a mongodb collection, see https://docs.mongodb.com/manual/core/schema-validation/
+ * @param {string} collection - The collection to set the schema for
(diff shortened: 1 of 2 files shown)


evq:
this enforces schemas even during manual edits. using `Joi` can't give us that. I agree the syntax leaves much to be desired - the jsonSchema support that was added later is better but unfortunately not supported for our mongo version.

mrose17:
we can either upgrade to mongo 3.6 or use joi in the interim. the syntax in 3.4 is a "dead letter".

evq:
I'm all for switching to 3.6 but I don't agree that using Joi in the interim makes more sense than using this syntax until then

mrose17:
we use joi throughout all the code, and in may repos. it is well-understood. the pre 3.6-syntax would e a one-off.

evq:
the pre 3.6-syntax is standard mongo query syntax which we also make use of :)

mrose17:
ok, let's use the mongo syntax. however, can we structure the .code so that the schema is an optional property for each object in the array passed to checkIndices, so that checkIndices can do the test if the property is present on any of them?

evq:
OBE

PR #182, opened 2018-04-10
referral payout report

evq:


eyeshade/controllers/reports.js
```diff
@@ -63,7 +63,8 @@ v3.publishers = {}
 v1.publisher.contributions = {
   handler: (runtime) => {
     return async (request, reply) => {
-      const authority = request.auth.credentials.provider + ':' + request.auth.credentials.profile.username
+      // const authority = request.auth.credentials.provider + ':' + request.auth.credentials.profile.username
```
mrose17: i think that `git` is able to keep track of what gets added/deleted/changed, so i'd much rather not see lines commented out throughout this commit...

eyeshade/workers/reports.js
```diff
@@ -927,9 +924,10 @@ const findEligPublishers = async (debug, runtime, publishers) => {
  **/
 const prepareReferralPayout = async (debug, runtime, authority, reportId, thresholdProbi) => {
   const owners = runtime.database.get('owners', debug)
+  const publishers = runtime.database.get('publishers', debug)
 
   const statements = await referralStatement(debug, runtime, undefined, true)
-  console.log(statements)
+  console.log(JSON.stringify(statements))
```
mrose17: could e put a comment before this, perhaps something like:

        // NB: REMOVE THIS SOON!

eyeshade/workers/reports.js
```diff
@@ -910,7 +910,7 @@ const referralStatement = async (debug, runtime, owner, summaryP) => {
 const findEligPublishers = async (debug, runtime, publishers) => {
   const publishersC = runtime.database.get('publishers', debug)
 
-  if (!publishers) {
+  if (!publishers || publishers.length === 0) {
```
mrose17: fyi: i don't think that it is possible for the `pubishers` argument to ever be null/zero/undefined/etc., but empty is certainly possible...regardless,  it is good to have the check in the way you have it, as a safety.

eyeshade/controllers/reports.js
```diff
@@ -590,20 +631,21 @@ module.exports.routes = [
...
   await runtime.queue.create('report-surveyors-contributions')
   await runtime.queue.create('report-grants-outstanding')
 }
 
-function authorityProvider(request) {
-  const { auth } = request;
-  const { credentials } = auth;
-  const { provider, profile, } = credentials;
-  let authority = provider;
+function authorityProvider (request) {
```
mrose17: for consistency, this should be `const` not `function` ...

eyeshade/controllers/publishers.js
```diff
@@ -109,12 +109,9 @@ v2.settlement = {
         publisher = await publishers.findOne({ publisher: entry.publisher })
         if (!publisher) return reply(boom.badData('no such entry: ' + entry.publisher))
 
-        if (!publisher.owner) return reply(boom.badData('no owner for publisher: ' + entry.publisher))
-
-        owner = await owners.findOne({ owner: publisher.owner })
+        // The owner at the time of uploading could be different
```
mrose17: right, but why isn't the previous owner getting the settlement? if we're going to fix this, we really ought to take each vote with the owner at the time the vote is recorded.  so, if owner X sells publisher Z to owner Y on such-and-such a date, then votes cast upto that time, should go to owner X, not owner Y, irregardless of when the settlement report is generated.

the change above is probably no more wrong than the existing code, but it is not fixing the actual problem...

eyeshade/workers/reports.js
```diff
@@ -17,6 +17,7 @@ let altcurrency
 
 const datefmt = 'yyyymmdd-HHMMss'
 const datefmt2 = 'yyyymmdd-HHMMss-l'
+const feePercent = 0.05
```
mrose17: it's good to remove the constant from the running code... but isn't it better to get it from the config and have `initialize` set it with something like

        feePercent = runtime.config.processingFee || 0.05

eyeshade/workers/reports.js
```diff
@@ -815,6 +816,153 @@ const date2objectId = (iso8601, ceilP) => {
                        (ceilP ? 'ffffffffffffffff' : '0000000000000000'))
 }
 
+/**
+ * A referral statement consists of entries describing accrued earnings from referrals,
```
mrose17: technically, the accounting term "accrued earnings" is probably correct here (i'm not an accountant); however, i'm a little concerned about using accounting terms-of-art in code with an actual accountant reviewing it. i prefer to use generic terms to avoid any confusion by domain experts...

eyeshade/workers/reports.js
```diff
@@ -815,6 +816,153 @@ const date2objectId = (iso8601, ceilP) => {
...
+      $group: {
+        _id: '$publisher',
+        count: { $sum: 1 },
+        probi: { $sum: '$probi' }
+      }
+    }
+  ])
+  referralTotals.forEach((total) => {
+    total.publisher = total._id
+    total.probi = new BigNumber(total.probi)
```
mrose17: it is probably safer to use `...toString()` when taking a `Decimal128` to create a `BigNumber` -- if the argument to `BigNumber` is already a string, then adding a `.toString()` has, at most, a negligible performance impact. in this case, i think the argument isn't going to be a string.

eyeshade/workers/reports.js
```diff
@@ -815,6 +816,153 @@ const date2objectId = (iso8601, ceilP) => {
...
+    let balance = new BigNumber(0)
+    if (statements[publisher].referrals.summary.probi) {
+      balance = balance.plus(statements[publisher].referrals.summary.probi)
+      pubOwner = statements[publisher]
+    }
+    statements[publisher].settlements.summaries.forEach((summary) => {
+      balance = balance.minus(summary.probi)
+    })
+    if (balance.lessThan(0)) {
+      throw new Error('Publisher has been overpaid')
```
mrose17: if we're going to throw an error, at the very least, we should indicate *which* publisher; however, i wouldn't throw an error, i'd simply set it to zero and leave it at that. there may be legitimate reasons why past settlements exceeds contributions for a time. it is, of course, undesirable, but it's not clear to me that it's fatal. we could, of course, include a notification to the `eyeshade-bot` if the balance goes negative, so as to alert someone to see what's what... and, of course, please include the publisher-identity that is so afflicated.

eyeshade/workers/reports.js
```diff
@@ -815,6 +816,153 @@ const date2objectId = (iso8601, ceilP) => {
...
+    payment.authority = authority
+    payment.transactionId = reportId
+
+    const entry = await owners.findOne({ owner: payment.owner })
+    if ((!entry) || (!entry.provider) || (!entry.parameters)) {
+      await notification(debug, runtime, entry.owner, payment.publisher, { type: 'verified_no_wallet' })
+      continue
+    }
+
+    const wallet = await runtime.wallet.status(entry)
```
mrose17: the call to `runtime.wallet.status` should be in a `try-catch` in case the underlying call to the uphold API fails... e.g.,

        } catch (ex) {
            await notification(debug, runtime, entry.owner, payment.publisher, { type: 'verified_invalid_wallet' })
        }

eyeshade/workers/reports.js
```diff
@@ -843,6 +991,52 @@ exports.create = create
...
+      const reportId = payload.reportId
+      const summary = payload.summary
+      const thresholdProbi = payload.threshold || 0
+      const verified = payload.verified
+
+      if ((!balance) || (!summary) || (!authorized) || (!verified)) {
+        throw new Error('only summary && balance && authorized && verified is supported')
+      }
+
+      if (format === 'csv') {
```
mrose17: it is probably better to use `if (format !== 'json')`; regardless, i think you may find that being able to get a "csv" formatted file is nice, even if we don't use it as input to the settlement submission tool!

eyeshade/controllers/reports.js
```diff
@@ -55,6 +55,57 @@ v1.publishers = {}
...
+    scope: [ 'ledger', 'QA' ],
+    mode: 'required'
+  },
+
+  description: 'Returns information about referrals to publishers, used to prepare referral payout',
+  tags: [ 'api' ],
+
+  validate: {
+    query: {
+      format: Joi.string().valid('json', 'csv').optional().default('json').description('the format of the report'),
```
mrose17: since you error on requests for 'csv' reports, why not remove the choice of `, 'csv'" ?

eyeshade/workers/reports.js
evq: somewhat, but this is a `const` whereas I think following that pattern allows for accidental reassignment and I don't think we'll change the processing fee soon

eyeshade/workers/reports.js
evq: :+1:

eyeshade/controllers/publishers.js
evq: the owner in the settlement file is the one that was actually paid - this change made it so the settlement entry in db always reflected that

eyeshade/workers/reports.js
```diff
@@ -946,20 +945,24 @@ const prepareReferralPayout = async (debug, runtime, authority, reportId, thresh
...
 
-    payment.address = wallet.address
-    payment.currency = wallet.defaultCurrency
+      payment.address = wallet.address
+      payment.currency = wallet.defaultCurrency
 
-    payments.push(payment)
+      payments.push(payment)
+    } catch (ex) {
+      await notification(debug, runtime, payment.owner, payment.publisher, { type: 'verified_no_wallet' })
```
mrose17: i think you want `s/verified_no_wallet/verified_invalid_wallet/`

eyeshade/workers/reports.js
```diff
@@ -855,7 +855,7 @@ const referralStatement = async (debug, runtime, owner, summaryP) => {
   ])
   referralTotals.forEach((total) => {
     total.publisher = total._id
-    total.probi = new BigNumber(total.probi)
+    total.probi = new BigNumber(total.probi.toString())
```
mrose17: i think you'll need the same fix at line 890:

        summary.probi = new BigNumber(summary.probi)

PR #183, opened 2018-04-10
removes bitcoin and other legacy routes

### LINKED ISSUE #65: remove v1 legacy api support (bitcoin)

maikelmclauflin:
reverted removal of route params and valid cases from joi validation on params
fixes #65


mmclau14:
@evq we should be good here.

PR #187, opened 2018-04-11
Use the group aggregator to reduce all the settlements for the GET /v1/owners/.../wallet endpoint

mrose17:


eyeshade/controllers/owners.js
```diff
@@ -464,15 +464,47 @@ v1.getWallet = {
         }
       }
 
-      entries = await settlements.find({ owner: owner }, { sort: { timestamp: -1 }, limit: 1 })
-      entry = entries && entries[0]
+      entries = await publishers.find({ owner: owner })
+      summary = await settlements.group(
```
evq: https://docs.mongodb.com/manual/reference/method/db.collection.group/
> Deprecated since version 3.4: Mongodb 3.4

We are running version 3.4.9 in prod I think

mrose17: _typical. just typical._ i now have to find what the equivalent command is.

mrose17:
by the way: _deprecated_ doesn't mean it's gone, just that it will go away some day.

evq:
sure - it just means it shouldn't be used in new code

evq:
FWIW I think with some tweaks something like:
```
settlements.aggregate(
   [ { $match: { owner: owner } },
     { $sort: { timestamp: 1 } },
     {
       $group:
         {
           _id: { publisher: '$publisher', type: '$type' },
           timestamp: { $last: "$timestamp" }
           amount: { $last: "$amount" }
           probi: { $last: "$probi" }
           fees: { $last: "$fees" }
           commission: { $last: "$commission" }
         }
     }
   ]
)
```
can get you the latest settlement transaction for each publisher / settlement type. Then can easily sum with a grouping condition timestamp.

mrose17:
thanks. that's pretty close. take a look at what we have now...

PR #188, opened 2018-04-11
Use normalized publisher when doing verified lookup

mrose17:
Fixes https://github.com/brave-intl/bat-publisher/issues/28


mrose17:
@mmclau14 - please see if it makes sense to do a test case for this and let me know what you think. thanks!

PR #193, opened 2018-04-12
Typescript would have been helpful...

mrose17:
…I suppose

### DIFF
--- a/eyeshade/workers/reports.js
+++ b/eyeshade/workers/reports.js
@@ -1324,7 +1324,7 @@ exports.workers = {
       if (publisher) {
         query = { publisher: publisher }
         entries = await settlements.find(query)
-        publishers = await mixer(debug, runtime, publisher, range)
+        publishers = await mixer(debug, runtime, publisher && [ publisher ], range)
       } else {
         query = underscore.pick(payload, [ 'owner', 'hash', 'settlementId' ])
         if (range) query._id = range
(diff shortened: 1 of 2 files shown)


eyeshade/workers/reports.js
```diff
@@ -1324,7 +1324,7 @@ exports.workers = {
       if (publisher) {
         query = { publisher: publisher }
         entries = await settlements.find(query)
-        publishers = await mixer(debug, runtime, publisher, range)
+        publishers = await mixer(debug, runtime, publisher && [ publisher ], range)
```
evq: aren't we checking if publisher is truthy above? so just `[ publisher ]` would be sufficient?

mrose17: in javascript, the last non-null value is the result, so `publisher && [ publisher ]` is equivalent to `publisher ? [ publisher ] : undefined` (or perhaps `null` or `0`)

evq: sure, but as I said we know publisher is truthy because we do this directly above:
```
if (publisher) {
```
so what you wrote here is:
```
true ? [ publisher ] : undefined
```
which obviously makes more sense to write as:
```
[ publisher ]
```

mrose17: at times i must be blind (in this case, not seeing the `if (publisher) {` ... cf. https://github.com/brave-intl/bat-ledger/pull/194

PR #201, opened 2018-04-18
publisher / owner statement includes referral information, note field

evq:


eyeshade/workers/reports.js
```diff
@@ -1377,7 +1421,7 @@ exports.workers = {
       })
       if (!publisher) {
         data.push({
-          publisher: 'TOTAL',
+          note: 'TOTAL IN',
```
mrose17: i suspect that we'll get an ask to change /IN/ to something else, but i'm not sure what.

eyeshade/workers/reports.js
```diff
@@ -1386,7 +1430,7 @@ exports.workers = {
         })
         if ((!summaryP) && (!payload.owner)) data.push([])
         data.push({
-          publisher: 'TOTAL',
+          note: 'TOTAL OUT',
```
mrose17: ditto the comment above, though i suspect "OUT" will be replaced with "SETTLEMENTS?

evq:
:airplane: :rocket:

Issue #174, opened 2018-04-05
create full lifecycle for grant server during tests


mmclau14:
during the test for settling a grant, the tester should not have to manage the db before running the tests. manual management can cause the db to be in poor state and for tests to fail. instead, automate by making sure that during a test, a single grant is added to the db, it is used, and subsequently flushed from the redis db by the end of the test.

PR #211, opened 2018-04-27
publisher / owner balance calculation doesn't account for fees  - Issue 196

maikelmclauflin:


mmclau14:
fixes #174
npm scripts added for wiping the db

eyeshade/workers/reports.js
```diff
@@ -966,7 +966,12 @@ const prepareReferralPayout = async (debug, runtime, authority, reportId, thresh
 
     payment.owner = eligPublishers[i].owner
 
-    const entry = eligPublishers[i].ownerdata
+    const entries = eligPublishers[i].ownerdata
+    if (entries.length > 1) {
+      await notification(debug, runtime, payment.owner, payment.publisher, { type: 'verified_bad_ownerlink' })
```
evq: notification also notifies publishers, we should check what will happen if we send it a notification type it doesn't know about

mmclau14:
good for a check @evq

eyeshade/workers/reports.js
```diff
@@ -957,38 +957,52 @@ const prepareReferralPayout = async (debug, runtime, authority, reportId, thresh
...
   for (let i = 0; i < eligPublishers.length; i++) {
-    const payment = statements[eligPublishers[i].publisher].balance
+    const publisher = eligPublishers[i].publisher
+    const payment = statements[publisher].balance
+    console.log(payment)
     payment.type = 'referral'
-    payment.fees = payment.probi.times(feePercent).truncated()
-    payment.probi = payment.probi.minus(payment.fees)
+    payment.fees = '0'
+    // payment.probi = payment.probi
```
evq: can you remove this commented out assignment? also, I think we may want payment.fees to be a bignumber instead of a string. it probably doesn't matter because we call toString down the line but best to keep the types consistent.

eyeshade/workers/reports.js
```diff
@@ -1181,7 +1200,7 @@ exports.workers = {
             provider = entry && entry.provider
             if (provider && entry.parameters) wallet = await runtime.wallet.status(entry)
 
-            if ((!wallet) || (!wallet.address) || (!wallet.defaultCurrency)) {
+            if (!includeUnpayable && (!wallet || !wallet.address || !wallet.defaultCurrency)) {
```
evq: it would be nice if we can consolidate this with the reworked logic you did above using `validateWallet`, not critical though

test/ledger.integration.test.js
```diff
@@ -391,18 +454,262 @@ test('integration : v2 grant contribution workflow with uphold BAT wallet', asyn
...
+  const res2 = await fetchReport({
+    domain,
+    reportId
+  })
+  const {
+    // text: body,
+    status
+  } = res2
+  t.is(status, 200)
+  // const json = body.split('\n').map(row => row.split(',').map(cell => {
```
evq: is this report now using json / the includeUnpayable option?

test/ledger.integration.test.js
```diff
@@ -391,18 +454,262 @@ test('integration : v2 grant contribution workflow with uphold BAT wallet', asyn
...
+  //   const last = cell.length - 1
+  //   const first = quote[cell[0]]
+  //   if (first && first === quote[cell[last]]) {
+  //     return cell.slice(1, last)
+  //   } else {
+  //     return cell
+  //   }
+  // }))
+  // console.log('contribution data', reportId, json)
```
evq: we're not checking that the contribution report is correct here right? might be a good TODO

test/ledger.integration.test.js
mmclau14: ah, this was old code. cleaned up. should be happening in the next test

PR #215, opened 2018-05-02
add more alt currency support

mrose17:
ready: BCH, BTG, DASH
waiting on uphold: XRP


bat-utils/lib/runtime-wallet.js
```diff
@@ -236,36 +243,81 @@ Wallet.prototype.purchaseBAT = async function (info, amount, currency, language)
...
+
+    if (this.runtime.config.currency.altcoins.indexOf(altcoin) === -1) return ('unconfigured altcoin: ' + altcoin)
+
+    cardInfo = await this.uphold.getCard(cardId)
+    addresses = (cardInfo && cardInfo.address) || {}
+    if (addresses[network]) return { id: addresses[network], network: network }
+
+    return this.uphold.createCardAddress(cardId, network)
+  },
+  addAddress: async function (info, altcoin) {
```
evq: can we consolidate `addAddress` and `createAddress`? 

it seems like the only difference is `addAddress` sets `info.addresses[altcoin]` - maybe we could remove from `create` and just use the consolidated function

evq:
@mrose17 can you add integration tests for the expected workflows? e.g. existing wallet needs to create uphold addresses for missing alt-currencies, new wallet creates all, etc?

bat-utils/lib/runtime-wallet.js
mrose17: short answer: no, because addAddress is called both internally and from the upper-layer...

evq: sorry I don't understand. if we put the functionality from `createAddress` in `addAddress` can't we continue to call it from internally and in the upper layer?

mrose17: the upper layer calls both `addAddress` and `create`. both of these call `createAddress`, which wants a `cardId` as a parameter. `create` could pass the `cardId`; however `addAddress` being called by the upper layer is given an `info` object, in which `info.providerId` is the `cardId`.  with the exception of `create` which returns an `info` object, all calls from the upper layer provider an `info` object. having  the upper layer call `address` with a `cardId` rather than an `info` object, would violate the encapsulation being used.

the current code takes the maximal amount of code needed for both `addAddress` and `create` and puts it in `createAddress`, thereby allow maximal code reuse...

evq: it seems like we have `info` in both places (being that `create` is the originator of the `info` - being it's return value)

evq:
I cleaned up the commit history by cherry-picking onto master

mikel2000:
XRP seems to be available now too: https://www.coindesk.com/payments-platform-uphold-completes-xrp-ledger-integration/.

PR #220, opened 2018-05-02
more database indices

mrose17:
Based on reports about slow queries


evq:
LGTM pending tests pass

Issue #225, opened 2018-05-04
Fix invalid error check


maikelmclauflin:
there is an invalid error check that occurs on ledger/controllers/registrar.js:164

PR #226, opened 2018-05-04
updated error check issue-225

maikelmclauflin:
basically, this is a bug in the code that does not send the correct response when a `httpSignature` is sent with a bad payload.

### DIFF
--- a/ledger/controllers/registrar.js
+++ b/ledger/controllers/registrar.js
@@ -161,7 +161,7 @@ const createPersona = function (runtime, apiVersion) {
         octets: Joi.string().optional().description('octet string that was signed and digested')
       }).required()
     }
-    var validity = Joi.validate(request.payload, requestSchema)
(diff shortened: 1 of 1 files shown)


ledger/controllers/registrar.js
```diff
@@ -161,7 +161,7 @@ const createPersona = function (runtime, apiVersion) {
         octets: Joi.string().optional().description('octet string that was signed and digested')
       }).required()
     }
-    var validity = (Joi.validate(request.payload, requestSchema).error)
+    var validity = Joi.validate(request.payload, requestSchema)
```
evq: nice find! do we want to validate `request.payload.request` instead of `request.payload`?

with this fix CI is failing with `"message": "child \\"headers\\" fails because [\\"headers\\" is required]"` - I see we use `request.payload.request.headers` below so I was thinking we might just be validating at the wrong level. thanks!

maikelmclauflin:
fixes #225

PR #231, opened 2018-05-07
publisher / owner balance calculation doesn't account for fees - issue-196

maikelmclauflin:


maikelmclauflin:
is this pr waiting for `includeUnpayable` to be removed from all statements functions in `controllers/reports.js`? if so, all, or which lines in particular?

evq:
yes. please remove includeUnpayable from all reports except the contributions and referral report endpoints / worker impls

PR #235, opened 2018-05-08
removed bitcoin support

maikelmclauflin:
rebased on current master, mostly just removing lines from codebase having to do with btc

### DIFF
--- a/bat-utils/lib/runtime-wallet.js
+++ b/bat-utils/lib/runtime-wallet.js
@@ -1,7 +1,6 @@
 const BigNumber = require('bignumber.js')
 const SDebug = require('sdebug')
 const UpholdSDK = require('@uphold/uphold-sdk-javascript')
-const bitcoinjs = require('bitcoinjs-lib')
 const crypto = require('crypto')
 const underscore = require('underscore')
 const { verify } = require('http-request-signature')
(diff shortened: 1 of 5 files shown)


maikelmclauflin:
fixes #65

test/ledger.integration.test.js
```diff
@@ -77,7 +77,6 @@ test('integration : v2 contribution workflow with uphold BAT wallet', async t =>
   t.true(response.body.hasOwnProperty('verification'))
 
   t.true(response.body.wallet.addresses.hasOwnProperty('BAT'))
-  t.true(response.body.wallet.addresses.hasOwnProperty('BTC'))
```
evq: these are deposit addresses on a card denominated in BAT.
we still want users to be able to deposit BTC which will be autoconverted by uphold to BAT, so we should still address that this address is created

bat-utils/lib/runtime-wallet.js
```diff
@@ -256,7 +228,6 @@ Wallet.providers.uphold = {
         }
         return { 'wallet': { 'addresses': {
           'BAT': ethAddr.id,
-          'BTC': btcAddr.id,
```
evq: see below

ledger/controllers/registrar.js
```diff
@@ -369,7 +342,6 @@ v2.createPersona =
       wallet: Joi.object().keys({
         paymentId: Joi.string().guid().required().description('opaque identifier for BTC address'),
         addresses: Joi.object().keys({
-          BTC: braveJoi.string().altcurrencyAddress('BTC').optional().description('BTC address'),
```
evq: see below

test/ledger.integration.test.js
maikelmclauflin: cool, i can put these marked lines back in.

PR #241, opened 2018-05-10
Fix various settlement report issues

mrose17:
In using reports (other than the report for the publishers site, some
issues came up: in particular:

- a report for a single publisher would always come back empty, due to
a missed variable.
- a report for a non-owner would fetch referrals
- a report for all publishers would process results for all publishers,
not just the ones with settlements


eyeshade/workers/reports.js
```diff
@@ -1356,49 +1372,52 @@ exports.workers = {
...
+          obj[item._id] = item
+          return obj
+        }, {})
+      } else referralTotals = {}
 
       publishers = underscore.keys(contributions)
-      if (payload.owner) {
-        publishers.push(...underscore.keys(referralTotals))
+      if (owner) {
+        publishers.push(underscore.keys(referralTotals))
```
evq: I think we want the spread operator that was removed here. We are grouping by publisher above and can potentially receive multiple elements back.

PR #254, opened 2018-05-16
POST /v2/owners adds verified site channels

### LINKED ISSUE #227: verify all channels sent to POST /v2/owners by publishers
currently only non-website channels are verified (all are sent by publishers), but with brave-intl/publishers#815 we will need this change to keep things in sync until we can remove eyeshade's copy of the channel/owner relationships

ayumi:
Fix #227 

Also adds tests.

cc @evq


ayumi:
closing in favor of #259

PR #255, opened 2018-05-16
switch from nsp to npm audit

### LINKED ISSUE #232: address all outstanding nsp check advisories
@evq @mrose17 we should revisit and come up with a game plan regarding when this should happen to reduce exposure

evq:
fixes #232 
also switch to stable node version for CI and a fork of the uphold SDK which has the version of `request` bumped


package.json
```diff
@@ -59,7 +59,7 @@
   "homepage": "https://github.com/brave-intl/bat-ledger",
   "dependencies": {
     "@octokit/rest": "^15.5.1",
-    "@uphold/uphold-sdk-javascript": "^2.0.4",
+    "@uphold/uphold-sdk-javascript": "github:brave-intl/uphold-sdk-javascript",
```
ayumi: this might be extra effort to maintain our fork.
our plan might be to switch back, and in that case an issue and comment explaining this would be nice

evq: opened #258 to track

PR #266, opened 2018-05-17
only allow contributions using the latest surveyor

### LINKED ISSUE #248: have payments always use the current contribution surveyor
see https://github.com/brave-intl/bat-ledger/blob/master/ledger/controllers/wallet.js#L244, we use the client provided surveyor. depending on client behavior perhaps we can reject payments where the surveyorId does not match that of the current surveyor

evq:
fixes #248


evq:
@mrose17 can you please confirm the client logic will successfully re-request the current surveyorId if an error is returned on call to the PUT endpoint used for payment?

it seemed that way based on the lines mentioned here: https://github.com/brave-intl/bat-ledger/issues/248#issuecomment-389747241

thanks!

mrose17:
short answer: it should "work", but it isn't optimal by any stretch of the imagination. there will be delays in between retries. if there was a particular error code to look for, we could optimize based on that.

evq:
@mrose17 great. since this should be an edge case for clients operating nominally, I think we can move forward as is

FYI as implemented we return HTTP error code 422 - please open an issue if there is another response code you would like

mrose17:
@evq - is 422 returned for any other situation?

evq:
@mrose17 it is also returned in case of a bad tx signature (we should also not retry), a legacy surveyor (also should not retry) and if the uphold transaction is returned with an invalid status.

PR #275, opened 2018-05-18
remove site verification

### LINKED ISSUE #217: Remove SiteChannel verification
To simplify Eyeshade and dedupe data, we decided to move the Site channel verification flow so it's entirely within Publishers (https://github.com/brave-intl/publishers/issues/815). The Publishers task and this one need to completed together.

As a result, Eyeshade would no longer have verification tokens or unverified channels.
(TODO: Check if unverified channels exist in Eyeshade).

evq:
now that publishers is handling site verification we can remove related code in eyeshade

fixes #217


evq:
this is currently blocked on https://github.com/brave-intl/publishers/issues/896 since the report used by stats depends on the tokens collection which is removed by this PR

PR #277, opened 2018-05-18
timestamp and identity handoff to publishers

maikelmclauflin:
updated timestamp and identity handlers to with env var feature
@brave-intl/publishers is still waiting on timestamp, but identity is there for now


ledger/controllers/publisher.js
```diff
@@ -484,21 +488,65 @@ v3.identity =
...
     }
   }
+
+  async function publishersHandleIdentity (request, reply) {
+    const debug = braveHapi.debug(module, request)
+    const {
+      query
+    } = request
+    const stringified = querystring.stringify(query)
+    const baseUrl = process.env.PUBLISHERS_URL
```
evq: we should probably read these out of the runtime config for consistency. to do so we will need to copy the `module.exports.publishers` block under eyeshade in `config.js` to under ledger

ledger/controllers/publisher.js
```diff
@@ -527,9 +614,9 @@ v3.timestamp =
   validate: {},
 
   response: {
-    schema: Joi.object().keys({
-      timestamp: Joi.string().regex(/^[0-9]+$/).required().description('an opaque, monotonically-increasing value')
-    }).unknown(true).description('information about the most recent publisher timestamp')
+    // schema: Joi.object().keys({
```
evq: I'm assuming this will get uncommented once the publishers timestamp endpoint is testable?

maikelmclauflin:
currently tests are blocked because of @brave-intl/publishers#886

PR #279, opened 2018-05-21
expose scope in GET owner wallet endpoint

evq:
currently publishers is not able to determine if it needs to ask for re-authorization in order to create a new wallet card - this exposes scope so it can do so

### DIFF
--- a/test/eyeshade/owners.integration.test.js
+++ b/test/eyeshade/owners.integration.test.js
@@ -123,9 +123,11 @@ test('eyeshade POST /v2/owners with site channels', async t => {
 })
 
 test('eyeshade PUT /v1/owners/{owner}/wallet with uphold parameters', async t => {
-  t.plan(15)
+  t.plan(16)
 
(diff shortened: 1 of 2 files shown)


test/eyeshade/owners.integration.test.js
```diff
@@ -186,6 +189,8 @@ test('eyeshade PUT /v1/owners/{owner}/wallet with uphold parameters', async t =>
   t.is(Array.isArray(wallet['possibleCurrencies']), true, 'get wallet returns currencies we could create a card for')
   t.is(wallet['possibleCurrencies'].indexOf('BAT') !== -1, true, 'wallet can have a BAT card')
```
maikelmclauflin: consider using dot notation, and deconstruct where possible. this is a great example of where it makes a lot of sense.

evq: :+1: we should switch this file over to that convention

PR #282, opened 2018-05-22
flush data from db at the end of tests

### LINKED ISSUE #237: Tests: Database should be cleaned between tests
To avoid unexpected interactions and allow use of fixtures, tests should clean some or all database data during setup or teardown.

maikelmclauflin:
fixes #237


test/e2e.integration.test.js
```diff
@@ -570,3 +574,31 @@ test('ensure referral balances are computed correctly', async t => {
   t.is(walletProbi.toString(), reportProbi)
   assertWithinBounds(t, amount, 5.00, 0.25, 'USD value for a referral should be approx $5')
 })
+
+test.after(async () => {
+  await Promise.all([
+    cleanEyeshadeDb([
```
evq: can we just delete all collections via `db.dropDatabase()`? seems like this list will be annoying to maintain

test/e2e.integration.test.js
```diff
@@ -570,3 +574,31 @@ test('ensure referral balances are computed correctly', async t => {
...
+      'owners',
+      'restricted',
+      'publishers',
+      'tokens',
+      'referrals',
+      'surveyors',
+      'settlements'
+    ]),
+    cleanLedgerDb([
+      'owners',
```
evq: here as well

.travis.yml
```diff
@@ -24,6 +24,7 @@ env:
     - "REDIS_URL=redis://redis:6379"
     - "TESTING_COHORTS=test"
     - "TOKEN_LIST=foobarfoobar"
+    - "GRANT_REDIS_HOST=localhost"
```
evq: can we add a default value that works within containers to `base.yml`

Issue #284, opened 2018-05-22
add maturity logic to surveyors


evq:
in line with https://github.com/brave-intl/bat-ledger/issues/267, once we freeze surveyors we should introduce logic that only counts frozen surveyors toward publisher balances

maikelmclauflin:
maturity occurs in the daily process

PR #286, opened 2018-05-23
added indices for surveyors and contributions added response booleans

### LINKED ISSUE #267: add logic to freeze surveyors
a frozen surveyor no longer accepts votes
1. we will retroactively freeze surveyors older than N days
2. we will add logic to freeze surveyors automatically after N days

maikelmclauflin:
fixes #267, #284


eyeshade/workers/reports.js
```diff
@@ -109,6 +115,49 @@ const daily = async (debug, runtime) => {
...
+  }
+  const allNonFrozenSurveyors = await surveyors.find(where)
+  const freezeUpdates = allNonFrozenSurveyors.map(surveyorExpirer)
+  await Promise.all(freezeUpdates)
+    .then(() => freezeUpdates.length)
+
+  function surveyorExpirer (surveyor) {
+    const {
+      _id,
+      timestamp
```
maikelmclauflin: don't use timestamp. instead use the id since it holds the creation time

eyeshade/workers/wallet.js
```diff
@@ -189,14 +193,18 @@ exports.workers = {
     async (debug, runtime, payload) => {
       const surveyorId = payload.surveyorId
       const surveyors = runtime.database.get('surveyors', debug)
-      let state
 
       payload.probi = bson.Decimal128.fromString(payload.probi.toString())
-      state = {
+      const $set = underscore.extend({
+        counts: 0,
+        mature: false,
```
evq: can we change mature / frozen to use `setOnInsert`. just as a precaution :)

test/utils.js
```diff
@@ -165,3 +174,18 @@ module.exports = {
   braveYoutubeOwner,
   braveYoutubePublisher
 }
+
+function createSurveyor () {
```
evq: I would prefer that we make probi and votes arguments

test/utils.js
```diff
@@ -147,7 +148,15 @@ const cleanRedisDb = async () => {
   })
 }
 
+const voteExchangeRate = 1 // 1 BAT per vote
```
evq: I would prefer we keep this in the e2e test. better to have the surveyor test just use it's own value so they're not interdependent

test/e2e.integration.test.js
```diff
@@ -22,29 +22,31 @@ import {
...
   t.plan(3)
   const url = '/v2/surveyor/contribution'
-  const data = { 'adFree': {
-    'fee': { 'USD': 5 },
-    'votes': suggestedVotes,
-    'altcurrency': 'BAT',
-    'probi': new BigNumber(suggestedVotes * voteExchangeRate).times('1e18').toString()
-  }}
-
+  const data = {
```
evq: should we use the new createSurveyor function?

ledger/controllers/surveyor.js
```diff
@@ -91,8 +91,16 @@ v2.read =
 
     surveyor = await server(request, reply, runtime)
     if (!surveyor) return
-
-    reply(underscore.extend({ payload: surveyor.payload }, surveyor.publicInfo()))
+    const {
```
evq: I think we can remove all ledger side changes, there isn't automatic propagation both ways so ledger will always report the surveyor is unfrozen

test/e2e.integration.test.js
maikelmclauflin: data and url are used in subsequent lines

ledger/controllers/surveyor.js
maikelmclauflin: gotcha

test/utils.js
maikelmclauflin: that's fine. i can scope another one there. is there a good default that can be used for this fn @evq ?

test/utils.js
maikelmclauflin: that makes sense.

test/e2e.integration.test.js
evq: I think we can just call createSurveyor twice.
```
createSurveyor()
id1 =     .get('/v2/surveyor/contribution/current')
createSurveyor()
id2 =     .get('/v2/surveyor/contribution/current')

File: eyeshade/workers/wallet.js (hunk 2 above)
Comment (maikelmclauflin) [reply eyeshade/workers/wallet.js]: that. makes a lot of sense

File: test/utils.js (hunk 4 above)
Comment (evq) [reply test/utils.js]: kinda depends what's under test.

e.g. in the surveyor test you could prob just use 1 vote and 1 probi - basically made up values since you aren't testing contributions and thus the exchange rate doesn't matter

File: eyeshade/workers/reports.js [hunk 7]
```diff
@@ -98,6 +103,10 @@ const daily = async (debug, runtime) => {
       await runtime.queue.send(debug, 'publisher-report',
                                underscore.pick(entry, [ 'owner', 'publisher', 'verified', 'visible' ]))
     }
+
+    const surveyorsCollection = database.get('surveyors', debug)
+
+    await freezeOldSurveyors(7, midnight, surveyorsCollection)
```
evq: it might be good to make 7 configurable via env var

eyeshade/workers/reports.js
```diff
@@ -109,6 +118,48 @@ const daily = async (debug, runtime) => {
...
+      timestamp
+    } = surveyor
+    const where = {
+      _id
+    }
+    if (!timestamp) {
+      // bad surveyor
+      return
+    }
+    const tsmp = +createdTimestamp(_id)
```
evq: is the `+` here intentional?

ledger/controllers/surveyor.js
```diff
@@ -422,7 +427,8 @@ const create = async (debug, runtime, surveyorType, payload, parentId) => {
       surveyorType: surveyorType,
       active: surveyorType !== 'contribution',
       available: true,
-      payload: payload
+      payload: payload,
+      frozen: false
```
evq: I think we can remove this and everything having to do with frozen from `ledger`, only `eyeshade` needs to be aware of frozen state

test/eyeshade/surveyor.integration.test.js
```diff
@@ -0,0 +1,97 @@
...
+  function querySurveyor () {
+    return eyeshade.collection('surveyors').findOne({
+      surveyorId
+    })
+  }
+
+  async function waitUntilPropagated (fn) {
+    let finished = await fn()
+    while (!finished) {
+      await timeout(4000)
```
evq: this seems a little long between polls - for testing I tend to err on polling slightly too often

eyeshade/workers/reports.js
```diff
@@ -109,6 +118,48 @@ const daily = async (debug, runtime) => {
   debug('daily', 'running again ' + moment(tomorrow).fromNow())
 }
 
+exports.freezeOldSurveyors = freezeOldSurveyors
+
+async function freezeOldSurveyors (olderThanDays, anchorTime, surveyors) {
+  // in seconds
+  const nDays = DAY_SECONDS * olderThanDays
```
evq: you might consider extracting the time comparison logic into a function that takes `olderThanDays`, `anchorTime` and a mongo id. prime target for some unit testing if you do so

eyeshade/workers/reports.js
```diff
@@ -109,6 +118,48 @@ const daily = async (debug, runtime) => {
...
+
+  function surveyorExpirer (surveyor) {
+    const {
+      _id,
+      timestamp
+    } = surveyor
+    const where = {
+      _id
+    }
+    if (!timestamp) {
```
evq: you can probably ditch this. assuming it doesn't hurt to set frozen on a bad surveyor

eyeshade/workers/reports.js
maikelmclauflin: that makes sense

eyeshade/workers/reports.js
maikelmclauflin: yes. it turns the value into a number. just did some tests and found that it is an unnecessary step as you can compare a date object to a number

maikelmclauflin: ... so i will remove

test/eyeshade/surveyor.integration.test.js
maikelmclauflin: generally, it takes 8-12 seconds. can shorten

eyeshade/workers/reports.js
maikelmclauflin: that's fair

eyeshade/workers/reports.js
maikelmclauflin: would you prefer this to default to 7? or should it err out if it is not given a date

PR #287, opened 2018-05-23
simplify wallet balance calc

evq:


ledger/controllers/wallet.js
```diff
@@ -67,13 +67,16 @@ const read = function (runtime, apiVersion) {
...
-        let number = new BigNumber(confirmed)
-        let memo = Promise.resolve(number)
-        let ifGrantExpired = (grant) => runtime.wallet.isGrantExpired(info, grant)
-        let markAsExpired = (grant) => runtime.wallet.expireGrant(info, wallet, grant)
-        let reducer = sumActiveGrants(ifGrantExpired, markAsExpired)
-        balances.confirmed = await wallet.grants.reduce(reducer, memo)
+        wallet.grants.forEach((grant) => {
+          if (grant.status === 'active') {
+            const grantContent = braveUtils.extractJws(grant.token)
+            if runtime.wallet.isGrantExpired(info, grant) {
```
maikelmclauflin: this line will err. 
i left this as it was in order to create a block of code that could be tested by itself. Instead of adding to the monolithic code, i decided to add a layer of abstraction in order to think about what it is doing and how separately.

PR #294, opened 2018-05-29
Adds batching for ballots prepare phase

NejcZdovc:


NejcZdovc:
@evq ok code is done. I will add test, but for some reason travis is failing. Can you please check why is that?

NejcZdovc:
closing in favour of #306

Issue #301, opened 2018-05-30
switch to npm ci for travis, dockerfile builds


evq:
https://docs.npmjs.com/cli/ci

- [x] travis
- [ ] dockerfile

maikelmclauflin:
my b. coming up

PR #315, opened 2018-06-04
updated travis to use npm ci

maikelmclauflin:
fixes half of #301


maikelmclauflin:
nooice @evq took ~1 minute off

PR #324, opened 2018-06-12
forwarding timestamp to publishers

### LINKED ISSUE #218: Proxy ledger publisher verification related endpoints to Publishers app
In the Brave browser, the payments tab has publisher verified green checkmarks. This green checkmark is queried from Ledger. 

Instead of Ledger fetching this information from Eyeshade, we're going to fetch the verified_status for an owner from the publishers api.

We'll be disabling the data propagation in a separate task.

maikelmclauflin:
fixes #218

### DIFF
--- a/config.js
+++ b/config.js
@@ -9,16 +9,6 @@ const services = {
       { BAT : process.env.BAT_SETTLEMENT_ADDRESS                || '0x7c31560552170ce96c4a7b018e93cddc19dc61b6'
       }
 
-      module.exports.publishers = {}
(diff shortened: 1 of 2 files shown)


config.js
```diff
@@ -46,12 +56,6 @@ const services = {
     portno: 3002,
 
     f: () => {
-      if (process.env.PUBLISHERS_URL) {
```
evq: I don't think we want to delete this. eyeshade uses this elsewhere

PR #333, opened 2018-06-18
added stats endpoint and tests

### LINKED ISSUE #321: need wallets listing
Hello

Stats used to have direct access to the eyeshade mongo instance. It no longer does, and so we need an endpoint that will give us results similar to [this 'wallets' query](https://github.com/brave/vault-analytics/blob/master/src/bat_eyeshade.js).

maikelmclauflin:
fixes #321


maikelmclauflin:
huh. i guess i just had a wrong env var?

.travis.yml
```diff
@@ -19,6 +19,7 @@ env:
     - "GRANT_REDIS_HOST=localhost"
     - "GRANT_SIGNATOR_PUBLIC_KEY=bbe028009e744422bfc39b0e226633324d8c5fe0744c87c1018b88a488bc5899"
     - "GRANT_WALLET_CARD_ID=f042845f-fa62-4022-8117-a476ec583a7b"
+    - "MONGODB_URI=mongodb://mongo/ledger"
```
evq: do we need this line?

ledger/controllers/wallet.js
```diff
@@ -468,12 +468,199 @@ v2.lookup = { handler: (runtime) => {
...
+          $cond: {
+            if: {
+              $gt: ['$paymentStamp', 0]
+            },
+            then: 1,
+            else: 0
+          }
+        },
+        grants: {
+          $ifNull: ['$grants', []]
```
evq: I think we care mostly about active grants, we may want to use the logic below checking status here

PR #342, opened 2018-06-22
updated bignumber config

### LINKED ISSUE #323: Remove alternative bignumber config setups
project should use single bignumber config so that we are dealing with the same constrictions at every level

maikelmclauflin:
fixes #323


bat-utils/lib/extras-utils.js
```diff
@@ -39,3 +57,11 @@ function createdTimestamp (id) {
 function uint8tohex (arr) {
   return [].slice.call(arr).map((b) => ('00' + b.toString(16)).substr(-2)).join('')
 }
+
+function toBat (probi) {
```
evq: this belongs better in `runtime-currency.js`, but since it isn't used I would suggest removing it

bat-utils/lib/extras-utils.js
```diff
@@ -1,13 +1,31 @@
...
-  documentOlderThan
+  documentOlderThan,
+  toBat,
+  mongoUri,
+  justDate
 }
 
 const DAY_MS = 60 * 60 * 24 * 1000
+function mongoUri (db) {
+  return `${process.env.BAT_MONGODB_URI}/${db}`
```
evq: This env var is test specific, we should not have it here

bat-utils/lib/extras-utils.js
```diff
@@ -39,3 +57,11 @@ function createdTimestamp (id) {
 function uint8tohex (arr) {
   return [].slice.call(arr).map((b) => ('00' + b.toString(16)).substr(-2)).join('')
 }
+
+function toBat (probi) {
+  return (new BigNumber(probi || 0)).dividedBy(1e18)
+}
+
+function justDate (date) {
```
evq: this isn't used, so let's remove it

bat-utils/lib/extras-utils.js
```diff
@@ -1,13 +1,31 @@
+// this can be abstracted elsewhere as soon as we finish #274
+const BigNumber = require('bignumber.js')
+
+const dotenv = require('dotenv')
```
evq: We can remove this dotenv since we're removing the mongouri stuff

bat-utils/lib/extras-utils.js
```diff
@@ -1,13 +1,31 @@
+// this can be abstracted elsewhere as soon as we finish #274
```
evq: maybe this belongs best in runtime currency?

PR #367, opened 2018-08-10
added minimumReconcileTimestamp to grant objects

### LINKED ISSUE #361: return customized minimum reconcile timestamp for paymentId
if at the time of claiming a new grant a wallet already has a grant which can cover their next contribution then due to the FIFO order of grant use we can avoid pushing back the users contribution date.

we could consider using the stored `unsignedTx` in order to make this determination

maikelmclauflin:
added minimumReconcileTimestamp to grant objects when unsigned transaction amount is less than the total of grants

fixes #361

### DIFF
--- a/ledger/controllers/grants.js
+++ b/ledger/controllers/grants.js
@@ -706,12 +706,15 @@ module.exports.initialize = async (debug, runtime) => {
         batchId: '',
         timestamp: bson.Timestamp.ZERO,
 
-        protocolVersion: 1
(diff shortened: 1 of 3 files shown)


maikelmclauflin:
changes should only be applied to `GET /vx/grants`

evq:
I don't understand what effect this will have. In any case I think this is OBE, we will likely do away with the `minimumReconcileTimestamp` mechanism

PR #369, opened 2018-08-13
Captcha bypass

### LINKED ISSUE #368: allow servers to bypass captcha with known token
servers that go through the same workflow without a ui would benefit from being able to skip the captcha

maikelmclauflin:
fixes #368

### DIFF
--- a/ledger/controllers/grants.js
+++ b/ledger/controllers/grants.js
@@ -362,9 +362,9 @@ v1.write = { handler: (runtime) => {
 
   validate: {
     params: { paymentId: Joi.string().guid().required().description('identity of the wallet') },
-    headers: {
+    headers: Joi.object().keys({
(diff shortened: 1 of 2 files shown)


NejcZdovc:
can we publish this logic under new prefix, so that bl is not effected (for example under v4)

ledger/controllers/grants.js
```diff
@@ -262,7 +262,9 @@ v1.write = { handler: (runtime) => {
     wallet = await wallets.findOne({ paymentId: paymentId })
     if (!wallet) return reply(boom.notFound('no such wallet: ' + paymentId))
 
-    if (runtime.config.captcha) {
+    const configCaptcha = runtime.config.captcha
+    const bypassCaptcha = request.headers['bypass-captcha']
+    if (configCaptcha && bypassCaptcha !== configCaptcha.bypass) {
```
evq: we should check node_env

PR #376, opened 2018-08-17
Send version to sentry

maikelmclauflin:
fixes #376


evq:
what do you think about using the commit sha provided by this instead? - https://devcenter.heroku.com/articles/dyno-metadata

PR #383, opened 2018-08-29
Table balances endpoint hotfix

maikelmclauflin:


eyeshade/controllers/accounts.js
```diff
@@ -78,7 +78,7 @@ v1.getBalances =
 { handler: (runtime) => {
   return async (request, reply) => {
     let accounts = request.query.account
-    if (!accounts) return reply(boom.badData)
+    if (!accounts) return reply(boom.badData())
```
nvonpentz: @maikelmclauflin what does the endpoint return in this situation currently?

We're going to be using this endpoint to lookup balances for unverified publishers which there might not be existing accounts for.  I think we'd want to return a empty response or 0 balances for those

maikelmclauflin: @nvonpentz i don't think this endpoint currently allows for this line to be hit because of the joi validation below. i'll remove.

evq: @nvonpentz the validation here is that some accounts were passed, not that those accounts exist - so we should be fine (doubly so since due to Joi validation this shouldn't get hit as Michael noted)

PR #386, opened 2018-09-11
Table balances endpoint hotfix

maikelmclauflin:
adds order and fixes 500s

### DIFF
--- a/eyeshade/controllers/accounts.js
+++ b/eyeshade/controllers/accounts.js
@@ -5,6 +5,8 @@ const utils = require('bat-utils')
 const braveHapi = utils.extras.hapi
 const braveJoi = utils.extras.joi
 
+const queries = require('../lib/queries')
+
 const v1 = {}
 
 const orderParam = Joi.string().valid('asc', 'desc').optional().default('desc').description('order')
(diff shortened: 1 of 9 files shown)


eyeshade/controllers/accounts.js
```diff
@@ -149,10 +148,9 @@ v1.getEarningsTotals =
  from account_transactions
  where account_type = 'owner' and transaction_type = $1
  group by (account_id, channel)
- order by earnings $2
- limit $3;`
-
-    const amounts = await runtime.postgres.pool.query(query1, [type, order.toLowerCase(), limit])
+ order by paid ${order.toUpperCase()}
```
evq: I would prefer that we don't use regular text substitution here

evq: this doesn't seem like it would work, there is no field `paid` here

eyeshade/controllers/accounts.js
```diff
@@ -121,18 +123,19 @@ v1.getBalances =
...
 { handler: (runtime) => {
   return async (request, reply) => {
+    const debug = braveHapi.debug(module, request)
     let { type } = request.params
     let {
       order,
       limit
     } = request.query
-
+    debug('query', request.query)
```
evq: is this duped by our query method which also logs?

maikelmclauflin: ah, yes, can fix

Issue #375, opened 2018-08-16
swap out runtime-currency for requests to helper server


maikelmclauflin:
instead of having all individual projects connecting to gdax and coinmarketcap, it would be better to have a single server that does this and sends cached rate information whenever such information is requested.

PR #388, opened 2018-09-14
Runtime currency swap

### LINKED ISSUE #359: abstract runtime-currency into separate project
runtime currency is really a whole project in and of itself and should be sharable across any services that need this information so that we do not have to send the data across the network. create a new node module, abstract the file into it, and write tests to make sure the project works as expected and reintegrate into bat-ledger

maikelmclauflin:
fixes #359, #375  puts module accessing 3rd parties in other service


.travis.yml
```diff
@@ -13,6 +13,7 @@ env:
     - "BAT_HELPER_SERVER=http://localhost:3004"
```
evq: I think we can remove helper here as well as in `base.yml` and in `.env.example`

base.yml
```diff
@@ -15,6 +15,7 @@ services:
       - "BAT_MONGODB_URI=mongodb://mongo"
       - "BAT_POSTGRES_URL=postgres://eyeshade:password@eyeshade-postgres/eyeshade"
       - "BAT_REDIS_URL=redis://redis:6379"
+      - "BAT_RATIOS_URL=https://bat-ratio.herokuapp.com"
```
evq: we should create a new `basicattentiontoken.org` subdomain, the existing heroku app should be converted to our staging instance and a new production instance should be provisioned

maikelmclauflin: correct, just checked in with the devops issue: https://github.com/brave/devops/issues/230 (which I would assume would be this) can update when informed it is available.

.travis.yml
maikelmclauflin: good catch

bat-utils/lib/runtime-currency.js
```diff
@@ -1,649 +1,189 @@
...
+        authorization: `Bearer ${accessToken}`,
+        'content-type': 'application/json'
       }
     }
+    try {
+      const body = await braveHapi.wreck.get(endpoint, options)
+      const dataString = body.toString()
+      const data = JSON.parse(dataString)
+      const { payload } = data
+      cache.set(endpoint, payload)
```
evq: shouldn't this be `cache.set(currency:${endpoint}, payload)` to match the get above?

bat-utils/lib/runtime-currency.test.js
```diff
@@ -0,0 +1,78 @@
...
+  const knownRateKeys = currency.knownRateKeys
+  t.plan(2 + knownRateKeys.length)
+  let result
+  result = await currency.rates('BAT')
+  t.true(_.isObject(result))
+  t.true(knownRateKeys.length > 1)
+  knownRateKeys.forEach((key) => {
+    t.true(_.isNumber(+result[key]))
+  })
+})
```
evq: can you add a test to ensure caching is working correctly?

eyeshade/workers/reports.js
```diff
@@ -1556,7 +1556,223 @@ exports.workers = {
       }
       runtime.notify(debug, { channel: '#publishers-bot', text: authority + ' report-publishers-statements completed' })
     },
+/* sent by GET /v1/reports/publishers/status
```
evq: this seems unrelated - maybe from a merge conflict?

bat-utils/lib/runtime-currency.test.js
maikelmclauflin: yis. added below

base.yml
maikelmclauflin: should be on terraform tomorrow. according to @brave-intl/devops

PR #395, opened 2018-09-25
Choices price shift

### LINKED ISSUE #390: automatically adjust server provided monthly budget based on xchange rate
Three asks per @mandar-brave and @davidtemkin  -

(Adjust based on exchange rate:)

1. Values on monthly budget menu
2. Default monthly budget (set to close to $5)
3. Value of one vote in BAT (calibrate to achieve ~100 votes per default monthly budget)

There is a privacy concern related to the third, namely that it results in surveyor rotation which should not occur too frequently.

maikelmclauflin:
fixes 1 and 2 in #390 
branch of #388

### DIFF
--- a/ledger/controllers/surveyor.js
+++ b/ledger/controllers/surveyor.js
@@ -4,9 +4,7 @@ const boom = require('boom')
 const bson = require('bson')
 const underscore = require('underscore')
 const surveyors = require('../lib/surveyor')
-const {
-  BigNumber
-} = require('bat-utils/lib/extras-utils')
+const BigNumber = require('bignumber.js')
 
 const utils = require('bat-utils')
 const braveHapi = utils.extras.hapi
--- a/ledger/controllers/surveyor.js
+++ b/ledger/controllers/surveyor.js
@@ -774,8 +774,8 @@ function choicesPrices (ratio) {
     [6, 10, 14, 20, 40],
     [9, 12, 20, 35, 50],
     [15, 25, 35, 50, 100],
-    [20, 35, 50, 85, 175],
-    [30, 50, 70, 100, 200]
+    [20, 35, 50, 85],
+    [30, 50, 70, 100]
   ]
   const priceIncrements = [1, 0.8, 0.6, 0.5, 0.35, 0.2, 0.15, 0.1]
   let index = underscore.findIndex(priceIncrements, (increment) => {
--- a/ledger/controllers/surveyor.js
+++ b/ledger/controllers/surveyor.js
@@ -295,23 +295,6 @@ v2.update =
 /*
    GET /v2/surveyor/{surveyorType}/{surveyorId}/{uId}
  */
-const currencyPattern = /^[0-9A-Z]{2,}$/
-const joiSurveyor = Joi.object().keys({
-  surveyorId: Joi.string().required().description('identifier for the surveyor'),
-  surveyVK: Joi.string().required().description('public key for the surveyor'),
-  registrarVK: Joi.string().required().description('public key for the associated registrar'),
-  signature: Joi.string().required().description('initialization response for the surveyor'),
-  payload: Joi.object().keys({
-    cohort: Joi.string().allow(['control', 'grant', 'test']),
-    adFree: Joi.object().keys({
-      fee: Joi.object().pattern(currencyPattern, Joi.number()).min(1).description('value after fees'),
-      choices: Joi.object().pattern(currencyPattern, Joi.array().items(Joi.number().required()).min(1)).description('currencies to choose to send'),
-      votes: Joi.number().optional().description('number of votes for distribution'),
-      altcurrency: Joi.string().regex(currencyPattern).optional().description('currency to be distributed'),
-      probi: Joi.string().regex(/\d+/).optional().description('amount of the altcurrency to be distributed')
-    }).required().description('settings for ad free mode')
-  }).required().description('additional information')
-}).required().description('response payload')
 
 v2.phase1 =
 { handler: (runtime) => {
@@ -361,12 +344,8 @@ v2.phase1 =
     })
 
     const payload = surveyor.payload
-    const surveyorJSON = underscore.extend({
-      signature,
-      payload
-    }, surveyor.publicInfo())
 
-    reply(surveyorJSON)
+    reply(underscore.extend({ signature: signature, payload: payload }, surveyor.publicInfo()))
   }
 },
 
@@ -383,7 +362,13 @@ v2.phase1 =
   },
 
   response: {
-    schema: joiSurveyor
+    schema: Joi.object().keys({
+      surveyorId: Joi.string().required().description('identifier for the surveyor'),
+      surveyVK: Joi.string().required().description('public key for the surveyor'),
+      registrarVK: Joi.string().required().description('public key for the associated registrar'),
+      signature: Joi.string().required().description('initialization response for the surveyor'),
+      payload: Joi.object().optional().description('additional information')
+    })
   }
 }
 
@@ -688,14 +673,15 @@ v2.batchSurveyor =
 
     const now = underscore.now()
 
-    const payload = surveyors.map((surveyor) => {
+    let payload = []
+    surveyors.forEach(surveyor => {
       signature = surveyor.sign(uId)
       runtime.newrelic.recordCustomEvent('sign', {
         surveyorId: surveyor.surveyorId,
         surveyorType: surveyor.surveyorType,
         duration: underscore.now() - now
       })
-      return underscore.extend({ signature, payload: surveyor.payload }, surveyor.publicInfo())
+      payload.push(underscore.extend({ signature, payload: surveyor.payload }, surveyor.publicInfo()))
     })
 
     reply(payload)
@@ -715,7 +701,13 @@ v2.batchSurveyor =
   response: {
     schema: Joi.alternatives().try(
       Joi.array().items(
-        joiSurveyor
+        Joi.object().keys({
+          surveyorId: Joi.string().required().description('identifier for the surveyor'),
+          surveyVK: Joi.string().required().description('public key for the surveyor'),
+          registrarVK: Joi.string().required().description('public key for the associated registrar'),
+          signature: Joi.string().required().description('initialization response for the surveyor'),
+          payload: Joi.object().optional().description('additional information')
+        })
       ),
       Joi.object().keys({
         statusCode: Joi.number().min(400).max(599).required(),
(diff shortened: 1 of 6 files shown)


evq:
blocked on #388

ledger/controllers/surveyor.js
```diff
@@ -4,6 +4,9 @@ const boom = require('boom')
 const bson = require('bson')
 const underscore = require('underscore')
 const surveyors = require('../lib/surveyor')
+const {
+  BigNumber
+} = require('bat-utils/lib/extras-utils')
```
nvonpentz: Do you know why extras-utils has BigNumber as a export?  It seems like everywhere else in the project BigNumber is required like this

```
const BigNumber = require('bignumber.js')
```

nvonpentz: It looks like the reason is so we can explicitly set the number of decimal places and when we want display exponential notation.  If this is the preferred way, should we change the other places where BigNumber is required?

maikelmclauflin: yeah, so i had an issue for this to create a clone of bignumber with a higher version number that supports scoped constructors. removed for now

ledger/controllers/surveyor.js
```diff
@@ -36,10 +37,45 @@ const server = async (request, reply, runtime) => {
     surveyor.payload = entry.payload
     surveyor.parentId = entry.parentId
   }
+  return surveyor && addChoices(runtime, surveyor)
+}
 
+async function addChoices (runtime, surveyor) {
+  const payload = surveyor.payload || {}
+  surveyor.payload = payload
```
nvonpentz: Do we need this line?  It looks like all the surveyors have a payload
```
db.surveyors.find({payload: {$exists: false}}).count()
> 0
```

ledger/controllers/surveyor.js
```diff
@@ -36,10 +37,45 @@ const server = async (request, reply, runtime) => {
...
   return surveyor
 }
 
+async function getAdjustedChoices (runtime, base, currencies) {
+  const rates = await runtime.currency.rates(base, currencies)
+  return underscore.mapObject(rates, choicesPrices)
+}
+
+function choicesPrices (ratio) {
+  const basePrice = (new BigNumber(1)).dividedBy(ratio)
```
nvonpentz: If we use BAT as the base in getAdjustedChoices can we remove this line here?

ledger/controllers/surveyor.js
```diff
@@ -36,10 +37,45 @@ const server = async (request, reply, runtime) => {
...
+  const table = [
+    [3, 5, 7, 10, 20],
+    [4, 6, 9, 12, 25],
+    [5, 8, 11, 17, 35],
+    [6, 10, 14, 20, 40],
+    [9, 12, 20, 35, 50],
+    [15, 25, 35, 50, 100],
+    [20, 35, 50, 85, 175],
+    [30, 50, 70, 100, 200]
```
evq: I think we won't have enough surveyors for some of these larger choices

ledger/controllers/surveyor.js
```diff
@@ -36,10 +37,45 @@ const server = async (request, reply, runtime) => {
...
+  ]
```
nvonpentz: We may want to run the values in the table by the product team before shipping if we haven't already

ledger/controllers/surveyor.js
nvonpentz: I'm assuming the rates that are returned are in terms of 1 unit of the base.  In this case since USD is the base, I would think that the rate would be how many BAT is 1 USD, right now about 6.4 BAT. 

So the basePrice as defined would be 1/6.4 or .156.  But that kind of confuses me since the price of 1 USD is not .156 BAT.  Where am I going wrong?

ledger/controllers/surveyor.js
```diff
@@ -259,6 +295,23 @@ v2.update =
...
+    cohort: Joi.string().allow(['control', 'grant', 'test']),
+    adFree: Joi.object().keys({
+      fee: Joi.object().pattern(currencyPattern, Joi.number()).min(1).description('value after fees'),
+      choices: Joi.object().pattern(currencyPattern, Joi.array().items(Joi.number().required()).min(1)).description('currencies to choose to send'),
+      votes: Joi.number().optional().description('number of votes for distribution'),
+      altcurrency: Joi.string().regex(currencyPattern).optional().description('currency to be distributed'),
+      probi: Joi.string().regex(/\d+/).optional().description('amount of the altcurrency to be distributed')
+    }).required().description('settings for ad free mode')
+  }).required().description('additional information')
+}).required().description('response payload')
```
nvonpentz: can we call this surveyorValidator for consistency with https://github.com/brave-intl/bat-ledger/blob/68e21c3ad269d376fb92aed8b807099619d3aa4d/ledger/controllers/grants.js#L53-L65?

ledger/controllers/surveyor.js
```diff
@@ -259,6 +295,23 @@ v2.update =
...
    GET /v2/surveyor/{surveyorType}/{surveyorId}/{uId}
  */
+const currencyPattern = /^[0-9A-Z]{2,}$/
+const joiSurveyor = Joi.object().keys({
+  surveyorId: Joi.string().required().description('identifier for the surveyor'),
+  surveyVK: Joi.string().required().description('public key for the surveyor'),
+  registrarVK: Joi.string().required().description('public key for the associated registrar'),
+  signature: Joi.string().required().description('initialization response for the surveyor'),
+  payload: Joi.object().keys({
+    cohort: Joi.string().allow(['control', 'grant', 'test']),
```
evq: this is overly strict

ledger/controllers/surveyor.js
evq: I would prefer we undo these validator changes and just constrain this PR to the exchange rate derived contribution amount changes

nvonpentz:
would it be possible to add a test ensuring the choices array is returned to the browser?

ledger/controllers/surveyor.js
maikelmclauflin: probably not it was just me being defensive i suppose

ledger/controllers/surveyor.js
maikelmclauflin: right so we're just trying to detect the closest approximated values and show the browser our estimate for how much they should be asking for (in usd) for it to be reasonable, or rather, the closest approximation of prices. 

I find that a good way to think about the ratios is as follows: how many Y can i get with X. With 1 USD, I can get 6.4 BAT. With 1 BAT I can get 0.156 USD. 

I decided to keep it simple and directly use the values that mandar had provided in his spreadsheet. This was so that if there were updates. I would not have to go through the step of transforming the data before putting it into the repo. I cannot currently find the spreadsheet, but I thought it would be better to just use the numbers from his point of view so that he could continue to think about it the same way.

ledger/controllers/surveyor.js
maikelmclauflin: agreed. they came from @mandar-brave perhaps he can paste the spreadsheet he provided here

ledger/controllers/surveyor.js
maikelmclauflin: oh, in the daily runner? i guess i just have to up the limit to, say 500 then?

ledger/controllers/surveyor.js
maikelmclauflin: fair enough

ledger/controllers/surveyor.js
maikelmclauflin: gotcha. then just a string?

ledger/controllers/surveyor.js
nvonpentz: Thanks @maikelmclauflin, that makes sense.  I guess my question is if USD is the base, then why is the basePrice .156 BAT.  Isn't the price of 1 USD 6.4 BAT?

maikelmclauflin: ah, yeah, after re-reading your quest, i guess we could. never thought of that. can modify

ledger/controllers/surveyor.js
nvonpentz: not a big deal, was just curious

ledger/controllers/surveyor.js
maikelmclauflin: https://docs.google.com/spreadsheets/d/1-X4lWEJBZLlYlJu9_2vhkx0PSjsC2OsBohNl7-N4d-c/edit#gid=0

ledger/controllers/surveyor.js
maikelmclauflin: is there anything that needs to be adjusted for this?

ledger/controllers/surveyor.js
nvonpentz: I feel like we should probably just define the relation between the price options so we could scale this list no matter what the price of BAT is.  We're kind of limiting ourselves to only the options in the hard coded table and we'd have to come back and add more options if prices went way up or way down.  Do you have thoughts on the matter?

I'm good with how you have it now since these are the approved values from the product team, and since it seems by this comment https://github.com/brave-intl/bat-ledger/pull/395#discussion_r260564322 we'll have a hard cap amount on the higher BAT amounts anyway.

ledger/controllers/surveyor.js
```diff
@@ -750,3 +751,38 @@ function getVoteRate () {
...
+  ]
+  const priceIncrements = [1, 0.8, 0.6, 0.5, 0.35, 0.2, 0.15, 0.1]
+  let index = underscore.findIndex(priceIncrements, (increment) => {
+    return increment <= ratio
+  })
+  if (underscore.isUndefined(index)) {
+    index = priceIncrements.length - 1
+  }
+  return table[index]
+}
```
nvonpentz: I may have found a bug here, i'm getting that any ratio under `.1` will return undefined instead of the highest option.

```
> choicesPrices(.1)
[ 30, 50, 70, 100 ]
> choicesPrices(.09)
undefined
```

If this is a valid bug, would you write a test case?

maikelmclauflin: for sure

ledger/controllers/surveyor.js
maikelmclauflin: yeah, i agree that we shouldn't be hard coding. in this case we're just kind of stuck between nice round values vs decimals and the estimations that round, man made values produce. I would like to move away from fiat as the base and move toward thinking about things in terms of BAT, even if the price does fluctuate, but that would take an effort from all areas of the business that i didn't think was worth delaying the feature over. I do think it's a discussion worth having. there are other companies that have stopped

PR #415, opened 2018-10-11
added endpoint for account ad viewing transactions

### LINKED ISSUE #413: add ads confirmation endpoint

maikelmclauflin:
fixes #413


eyeshade/controllers/accounts/index.js
```diff
@@ -0,0 +1,309 @@
...
+    } = request
+    const {
+      postgres
+    } = runtime
+    const parameters = {
+      paymentId: params.payment_id,
+      tokenId: params.token_id
+    }
+
+    const found = await transactions.findTransactionByToken(runtime, postgres, parameters)
```
evq: I think it's better to try to insert and handle the error case. (Since we have to handle the error case anyway since it's possible for two simultaneous requests to both go past this check)

eyeshade/controllers/accounts.js
```diff
@@ -1,250 +1 @@
...
-  }
-}
-
-module.exports.routes = [
-  braveHapi.routes.async().path('/v1/accounts/earnings/{type}/total').whitelist().config(v1.getEarningsTotals),
-  braveHapi.routes.async().path('/v1/accounts/settlements/{type}/total').whitelist().config(v1.getPaidTotals),
-  braveHapi.routes.async().path('/v1/accounts/balances').whitelist().config(v1.getBalances),
-  braveHapi.routes.async().path('/v1/accounts/{account}/transactions').whitelist().config(v1.getTransactions)
-]
+module.exports = require('./accounts/index.js')
```
evq: I would prefer we didn't nest this deeper in the folder hierarchy

eyeshade/lib/queries.js
```diff
@@ -37,3 +39,18 @@ function settlements (options = {}) {
...
+
+function findTransactionByToken () {
+  return `
+ select * from transactions
+ where transaction_type = 'ad_payout' and to_account = $1::text and document_id = $2::text
+ limit 1;
+`
+}
+
+function insertTransaction () {
```
evq: I think the sql queries for inserting transactions are distinct enough to justify simply in-lining them vs abstracting into a separate queries library

eyeshade/lib/transaction.js
```diff
@@ -5,13 +5,69 @@ const {
...
+}) {
+  const query = queries.insertTransaction()
+  const created = seconds()
+  const { config } = runtime
+  const id = uuidv5(`${paymentId}:${tokenId}`, '2ca02950-084f-475f-bac3-42a3c99dec95')
+  const probi = (new BigNumber(amount)).toString()
+  const month = monthsFromSeconds(created)
+  const replacements = [
+    id,
+    created,
```
evq: the `created_at` field can simply be omitted and postgres will use the current date as the default per https://github.com/brave-intl/bat-ledger/blob/master/eyeshade/migrations/0001_transactions/up.sql#L10

eyeshade/migrations/0001_transactions/up.sql
```diff
@@ -2,8 +2,8 @@ select execute($$
 
 insert into migrations (id, description) values ('0001', 'transactions');
 
-create type account_type as enum ('channel', 'owner', 'uphold', 'internal');
```
evq: these should be added in a new migrations file `eyeshade/migrations/0002_ads/up.sql` which modifies the existing enums

eyeshade/migrations/0001_transactions/up.sql
```diff
@@ -2,8 +2,8 @@ select execute($$
...
-create type transaction_type as enum ('contribution', 'referral', 'contribution_settlement', 'referral_settlement', 'fees', 'scaleup', 'manual');
+create type account_type as enum ('channel', 'owner', 'uphold', 'internal', 'paymentId');
+create type transaction_type as enum ('contribution', 'referral', 'contribution_settlement', 'referral_settlement', 'fees', 'scaleup', 'manual', 'ad_payout');
```
evq: we need two types, one into the `paymentId` account and one out. I suggest `ad` and `ad_settlement` for consistency (also `ad_payout` and `ad_settlement` would be confusing)

eyeshade/controllers/accounts/index.js
maikelmclauflin: that makes sense. can do

eyeshade/lib/transaction.js
maikelmclauflin: can remove.

eyeshade/controllers/accounts.js
```diff
@@ -242,9 +243,65 @@ v1.getPaidTotals =
...
+    const {
+      postgres
+    } = runtime
+    const {
+      amount: payloadAmount
+    } = payload
+    if (!_.isString(payloadAmount)) {
+      return reply(boom.badRequest())
+    }
+    const amount = (new BigNumber(payloadAmount)).dividedBy(1e18).toString()
```
evq: let's take amount not in probi - this matches the other accounts APIs

maikelmclauflin: fair enough.

PR #431, opened 2018-10-26
added auth layer for ads endpoints

maikelmclauflin:


eyeshade/controllers/accounts.js
```diff
@@ -100,7 +101,8 @@ v1.getBalances =
 },
 
   auth: {
-    strategy: 'simple',
+    strategy: 'ads',
+    scope: ['ads:read'],
```
evq: this will break access using the existing token that publishers uses right? I think we want to add a `publishers` scope that is also authed here that grants access to the existing token

maikelmclauflin: it would not, but i have updated to create own custom validation strategy.

PR #436, opened 2018-10-31
Mongo to postgres owner

maikelmclauflin:
added tests to start converting owners collection to postgres.


evq:
owners was migrated to publishers

PR #442, opened 2018-11-06
Update bat-utils/lib/runtime-currency.js

mrose17:
brave-core expects `Number` not `String` ... brave-core is being made more robust; however, in the interim...


aekeus:
I am approving this because of the impactful nature of the crash to our existing browser base. We will test this extensively on staging. After this goes to production, we will need to add tests @maikelmclauflin.

evq:
since we're sorted out the production issue (unrelated) - I am closing this in favor of #443 as I think it's a better fix

PR #454, opened 2018-11-14
Android attestation

maikelmclauflin:


ledger/controllers/grants.js
```diff
@@ -368,37 +488,77 @@ v1.claimGrant = { handler: (runtime) => {
...
+  const captchaHeaders = {
+    'Authorization': 'Bearer ' + captcha.access_token,
+    'Content-Type': 'application/json'
+  }
+  const wallets = database.get('wallets', debug)
+  const body = JSON.stringify({
+    token: headers['safetynet-token']
+  })
+
+  const payload = await braveHapi.wreck.post(url, {
```
evq: what error do we return if this fails? 5xx or 4xx?

it should be 5xx if there was a 5xx or timeout and 4xx if the response was a 4xx

ledger/controllers/grants.js
```diff
@@ -89,6 +98,36 @@ const getPromotions = (protocolVersion) => (runtime) => async (request, reply) =
...
+  const url = captcha.url + endpoint
+  const headers = {
+    'Authorization': 'Bearer ' + captcha.access_token,
+    'Content-Type': 'application/json'
+  }
+  const body = JSON.stringify({
+    token: request.headers['safetynet-token']
+  })
+
+  await braveHapi.wreck.post(url, {
```
evq: same question as below about error cases

ledger/controllers/grants.js
```diff
@@ -89,6 +98,36 @@ const getPromotions = (protocolVersion) => (runtime) => async (request, reply) =
   reply(entries)
 }
 
+/*
+ GET /v3/promotions
```
evq: we want this logic in `GET /v3/grants` instead of `GET /v3/promotions` since that's what the browser uses

ledger/controllers/grants.js
maikelmclauflin: throw ends up being the same thing as reply. statusCode is passed through via throw

ledger/controllers/grants.js
```diff
@@ -629,6 +789,7 @@ const getCaptcha = (protocolVersion) => (runtime) => {
     const { headers } = res
 
     const solution = JSON.parse(headers['captcha-solution'])
+    console.log('captcha solution', solution)
```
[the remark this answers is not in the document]
maikelmclauflin: thanks!

ledger/controllers/grants.js
```diff
@@ -271,26 +408,9 @@ v1.claimGrant = { handler: (runtime) => {
...
-      } else {
-        if (promotion.protocolVersion !== 1) {
-          return reply(boom.forbidden('must first request correct captcha version'))
-        }
-      }
-
-      if (!(checkBounds(wallet.captcha.x, captchaResponse.x, 5) && checkBounds(wallet.captcha.y, captchaResponse.y, 5))) {
-        return reply(boom.forbidden())
-      }
+    const invalidated = await validate(debug, runtime, request, promotion, wallet)
```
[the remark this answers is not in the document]
maikelmclauflin: makes sense

ledger/controllers/grants.js
```diff
@@ -252,10 +327,72 @@ const checkBounds = (v1, v2, tol) => {
...
+  validate: {
+    params: {
+      paymentId: Joi.string().guid().required().description('identity of the wallet')
+    },
+    headers: Joi.object().keys({
+      'safetynet-token': Joi.string().required().description('the safetynet token created by the android device')
+    }).unknown(true),
+    payload: Joi.object().keys({
+      promotionId: Joi.string().required().description('the promotion-identifier')
+    }).required().description('promotion derails')
```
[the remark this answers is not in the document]
maikelmclauflin: thanks

PR #467, opened 2018-12-18
expose wallet provider status and tweak check

evq:
per discussion with Uphold we should check if status === 'ok'.

for good measure we will expose via owner wallet info endpoint so admin dash can show


evq:
blocked by https://github.com/brave-intl/bat-publisher/pull/46

PR #468, opened 2018-12-18
Grant subset logic

### LINKED ISSUE #466: add logic in grant claim endpoints for ad claims
- new endpoint on ledger `GET /v3/grants` that will return an array of available promotions instead of single promotion as currently returned, addition of `type` field which can be `ugp` or `ads`.
- addition of type field above as well as optional `paymentId` field on objects in the `grants` collection, if present only advertised / claimable by the particular `paymentId`.
- if type field of claimed grant is `ads`, use the new grant instant redeem endpoint instead of associating the token with the wallet

maikelmclauflin:
fixes #466

### DIFF
--- a/ledger/controllers/grants.js
+++ b/ledger/controllers/grants.js
@@ -60,8 +60,9 @@ const promotionIdValidator = Joi.string().guid().description('the promotion-iden
 const altcurrencyValidator = braveJoi.string().altcurrencyCode().description('the grant altcurrency')
 const probiValidator = braveJoi.string().numeric().description('the grant amount in probi')
 const minimumReconcileTimestampValidator = Joi.number().description('time when the promotion can be reconciled')
-const grantValidator = Joi.string().description('the jws encoded grant')
-const grantsValidator = Joi.array().min(0).items(grantValidator).description('grants for bulk upload')
+const encodedGrantValidator = Joi.string().description('the jws encoded grant')
+const grantsValidator = Joi.array().min(0).items(encodedGrantValidator).description('grants for bulk upload')
+const expiryTimeValidator = Joi.number().positive().description('the time the grant expires')
 const grantProviderIdValidator = Joi.string().guid().when('type', {
   is: 'ads',
   then: Joi.required(),
@@ -72,21 +73,21 @@ const captchaResponseValidator = Joi.object().keys({
   x: Joi.number().required(),
   y: Joi.number().required()
 })
-const oldGrantValidator = Joi.object().keys({
+const grantContentValidator = Joi.object().keys({
   grantId: Joi.string().guid().required().description('the grant-identifier'),
   promotionId: promotionIdValidator.required(),
   altcurrency: altcurrencyValidator.required(),
   probi: probiValidator.required(),
   maturityTime: Joi.number().positive().required().description('the time the grant becomes redeemable'),
-  expiryTime: Joi.number().positive().required().description('the time the grant expires')
+  expiryTime: expiryTimeValidator.required()
 })
-const grantContentTypedValidator = oldGrantValidator.keys({
+const grantContentTypedValidator = grantContentValidator.keys({
   type: grantTypeValidator,
   providerId: grantProviderIdValidator
 })
 const publicGrantValidator = Joi.object().keys({
   altcurrency: altcurrencyValidator.optional().default('BAT'),
-  expiryTime: Joi.number().optional().description('the expiration time of the grant'),
+  expiryTime: expiryTimeValidator.optional(),
   probi: probiValidator.optional()
 }).unknown(true).description('grant properties')
 const publicGrantTypedValidator = publicGrantValidator.keys({
@@ -703,7 +704,7 @@ const uploadGrants = (protocolVersion) => (runtime) => {
     const promotionCounts = {}
     for (let entry of payload.grants) {
       const grantContent = braveUtils.extractJws(entry)
-      const validity = Joi.validate(grantContent, oldGrantValidator)
+      const validity = Joi.validate(grantContent, grantContentValidator)
       if (validity.error) {
         return reply(boom.badData(validity.error))
       }
--- a/ledger/controllers/grants.js
+++ b/ledger/controllers/grants.js
@@ -92,7 +92,6 @@ const v4 = {}
 
 /*
    GET /v2/promotions
-   GET /v3/promotions
  */
 
 const getPromotions = (protocolVersion) => (runtime) => async (request, reply) => {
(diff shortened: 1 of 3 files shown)


ledger/controllers/grants.js
```diff
@@ -880,3 +1083,215 @@ module.exports.initialize = async (debug, runtime) => {
...
+      }))
+    } catch (e) {
+      return reply(e)
+    }
+
+    return reply({
+      grants: result
+    })
+  }
+}
```
nvonpentz: Would you help me understand the purpose of this endpoint? It looks like we want to return the number of grants available for a given promotion type, is that right?

Why do we want to extend the query like this if the wallet has grants?
```
underscore.extend(query, { promotionId: { $nin: promotionIds } })
```

ledger/controllers/grants.js
```diff
@@ -303,6 +341,31 @@ v3.read = {
   }
 }
 
+/*
+  GET /v4/grants
+*/
+
+v4.read = {
+  handler: getGrantV4(4, chooseFromAvailablePromotionsByType),
```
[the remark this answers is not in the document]
maikelmclauflin: yeah, with so many different promotions often being added at different times, this is useful to isolate different parts and scope them / help isolate issues

ledger/controllers/grants.js
```diff
@@ -880,3 +1083,215 @@ module.exports.initialize = async (debug, runtime) => {
...
+  }) => {
+    return type === '' || type == null || type === 'ugp'
+  })
+  const adsPromotions = await chooseFromAvailablePromotions(debug, runtime, {
+    entries: ads
+  })
+  const ugpPromotions = await chooseFromAvailablePromotions(debug, runtime, {
+    entries: ugp
+  })
+  return [].concat(adsPromotions, ugpPromotions)
```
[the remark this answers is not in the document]
maikelmclauflin: so we want to return both in an array so that the client can see all the available grants available to it, from what i can tell perhaps @evq can give another reason

ledger/controllers/grants.js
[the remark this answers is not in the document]
nvonpentz: I was thinking that the handler function would already be isolated because of the name.  do we expect to have v5 routes to use the `getGrantV4` handler?

ledger/controllers/grants.js
```diff
@@ -303,6 +341,31 @@ v3.read = {
...
+  handler: getGrantV4(4, chooseFromAvailablePromotionsByType),
+  description: 'See if a v4 promotion is available',
+  tags: [ 'api' ],
+
+  validate: {
+    query: {
+      lang: Joi.string().regex(localeRegExp).optional().default('en').description('the l10n language'),
+      paymentId: Joi.string().guid().optional().description('identity of the wallet')
+    }
+  },
```
nvonpentz: I see we're using a `type` query parameter in the handler function https://github.com/brave-intl/bat-ledger/pull/468/files#diff-7fb1e89ea3f41276b410cb21071d95ddR1164, should we add validations for it?

ledger/controllers/grants.js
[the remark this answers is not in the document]
nvonpentz: so why does `getGrantV4` and `chooseFromAvailablePromotionsByType` take a `type` parameter if it's not being used?

ledger/controllers/grants.js
```diff
@@ -547,16 +650,31 @@ async function captchaCheck (debug, runtime, request, promotion, wallet) {
...
+}).required().description('data for bulk upload')
+const promotionV4Validator = promotionValidator.keys({
+  protocolVersion: protocolVersionValidator,
+  minimumReconcileTimestamp: Joi.number().optional().description('time when the promotion can be reconciled')
+})
+const promotionsV4Validator = Joi.array().min(0).items(promotionV4Validator)
+const grantsUploadV4Validator = grantsUploadSchema.keys({
+  grants: grantsValidator,
+  promotions: promotionsV4Validator
+})
```
nvonpentz: it looks like `grantsUploadV4Validator` and `grantsUploadSchema` are pretty similar.  is there difference in your mind between a validator and a schema? would it be possible to consolidate to using just one of those terms?

ledger/controllers/grants.js
```diff
@@ -616,6 +734,26 @@ const uploadGrants = (protocolVersion) => (runtime) => {
...
+  tags: [ 'api' ],
+
+  validate: { payload: grantsUploadSchema },
+
+  payload: { output: 'data', maxBytes: 1024 * 1024 * 20 },
+
+  response:
+    { schema: Joi.object().length(0) }
+}
+
```
nvonpentz: This endpoint was removed in https://github.com/brave-intl/bat-ledger/pull/505, was that a mistake?

ledger/controllers/grants.js
```diff
@@ -372,14 +435,50 @@ v3.claimGrant = {
...
+v4.claimGrant = {
+  handler: claimGrant(captchaCheck, v4CreateGrantQuery),
+  description: 'Request a grant for a wallet',
+  tags: [ 'api' ],
+
+  plugins: {
+    rateLimit: {
+      enabled: rateLimitEnabled,
+      rate: (request) => claimRate
+    }
```
nvonpentz: can we use the `rateLimitPlugin` here?

ledger/controllers/grants.js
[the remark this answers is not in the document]
maikelmclauflin: Oh, you're saying the name of the function that creates the closure should change. yeah. i can modify that

ledger/controllers/grants.js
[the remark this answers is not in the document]
maikelmclauflin: ah. nice. good catch

ledger/controllers/grants.js
maikelmclauflin: yeah, i should remove this

ledger/controllers/grants.js
maikelmclauflin: i think this was just because one existed before, so the suffix was just to differentiate between old and new. i suppose i can just add old to it to mark deprication

ledger/controllers/grants.js
maikelmclauflin: can remove

ledger/controllers/grants.js
maikelmclauflin: for sure

ledger/controllers/grants.js
maikelmclauflin: must have been a bad merge. can remove

ledger/controllers/grants.js
maikelmclauflin: yeah, i'll use that var

maikelmclauflin:
yeah, that makes sense. can do

PR #490, opened 2019-01-08
expose isMember from uphold user info

evq:


maikelmclauflin:
aha! i was just looking into this a bit ago.

PR #497, opened 2019-01-14
added top balances endpoint

### LINKED ISSUE #486: add API to expose top N channel balances

maikelmclauflin:
fixes #486

### DIFF
--- a/eyeshade/controllers/accounts.js
+++ b/eyeshade/controllers/accounts.js
@@ -78,7 +78,7 @@ ORDER BY created_at
 }
 
 /*
-   GET /v1/accounts/balances/top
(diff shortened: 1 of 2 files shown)


eyeshade/controllers/accounts.js
```diff
@@ -77,6 +77,54 @@ ORDER BY created_at
   }
 }
 
+/*
+   GET /v1/accounts/balances/top
```
evq: #486 is mostly concerned about channel balances, not owner or uphold.

How about making this `GET /v1/accounts/balances/{type}/top` so that publishers will end up calling `GET /v1/accounts/balances/channel/top`

PR #503, opened 2019-01-16
deny browser-laptop grants

### LINKED ISSUE #502: don't allow browser-laptop to claim grants
We didn't create a new API for the next round of grants, and as a result the old version of brave still has the ability to claim them.  In an effort to motivate users to upgrade their browser, we should prevent browser-laptop users to claim grants.  Specifically we should reject the `GET /v2/grants` request by looking at the user agent and rejecting all browsers running a chrome version < 70.

Additionally we should remove `GET /v1/grants` and `GET /v1/captchas/{payment_id}` since these are already no longer used.

I believe we will also want to remove these for the same reasons:
  * `GET /v1/promotions` 
  * `PUT /v1/grants/{paymentId}'`
  *  `POST /v1/grants`

nvonpentz:
Resolves #502 

Reject the requests to GET /v2/grants made by browsers running a chrome version < 70.

Remove
*  GET /v1/promotions
*  PUT /v1/grants/{paymentId}'
*  POST /v1/grants
*  GET /v1/captchas/{payment_id}
*  GET /v1/grants

### DIFF
--- a/ledger/controllers/grants.js
+++ b/ledger/controllers/grants.js
@@ -31,7 +31,6 @@ const joiGrantSubset = Joi.object().keys({
   probi: braveJoi.string().numeric().optional().description('the grant amount in probi')
 }).unknown(true).description('grant properties')
 
-const v1 = {}
 const v2 = {}
 const v3 = {}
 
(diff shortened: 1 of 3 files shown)


nvonpentz:
Superceded by #504

PR #507, opened 2019-01-17
Rate limit uphold

### LINKED ISSUE #476: Avoid hitting uphold API limits
In the past we've caused DOS'd ourselves by hitting our API limits on uphold, e.g. due to balance server issue having all balance queries poll through to uphold
a circuit breaker pattern there with max in-flight requests could be useful there

maikelmclauflin:
fixes #476

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ branches:
   - master
 env:
   global:
+    - "SERVICE=ledger"
     - "BAT_ADS_PAYOUT_ADDRESS=fb599883-c30c-4428-91be-ba901e69a0b2"
     - "BAT_BALANCE_SERVER=http://localhost:3003"
     - "BAT_EYESHADE_SERVER=http://localhost:3002"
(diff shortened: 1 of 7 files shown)


bat-utils/lib/extras-utils.js
```diff
@@ -2,13 +2,38 @@ const getPublisherProps = require('bat-publisher').getPublisherProps
...
+const ledgerLimiter = new Bottleneck({
+  reservoir: 50, // initial value
+  reservoirRefreshAmount: 25,
+  reservoirRefreshInterval: 10 * 1000, // must be divisible by 250
+  // also use maxConcurrent and/or minTime for safety
+  maxConcurrent: 32,
+  minTime: 0.1
+})
+
+const upholdLimiter = new Bottleneck({
```
evq: these limits are per instance - correct?

maikelmclauflin: yes. they would end up being per instance. i believe i can set it up to be shared across a redis store as well if that is preferrable

.travis.yml
```diff
@@ -6,6 +6,7 @@ branches:
   - master
 env:
   global:
+    - "SERVICE=ledger"
```
evq: hey why didn't we need this before?

maikelmclauflin: may be a relic from when i was debugging something. shall investigate

maikelmclauflin: this was so that utils tests that require configuration do not fail

PR #510, opened 2019-01-22
expose channel vote amounts

maikelmclauflin:
allows publishers to see pending channel votes

### DIFF
--- a/eyeshade/controllers/accounts.js
+++ b/eyeshade/controllers/accounts.js
@@ -1,7 +1,12 @@
 const Joi = require('joi')
+const { getPublisherProps } = require('bat-publisher')
 const boom = require('boom')
 const utils = require('bat-utils')
+const BigNumber = require('bignumber.js')
 const _ = require('underscore')
+const {
+  normalizeChannel
+} = require('bat-utils/lib/extras-utils')
 const queries = require('../lib/queries')
 const transactions = require('../lib/transaction')
 const braveHapi = utils.extras.hapi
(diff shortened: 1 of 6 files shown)


nvonpentz:
Is the idea here to have publishers hit both /v1/accounts/balances and /v1/accounts/balances/pending, sum the result, then display the sum to the user?

maikelmclauflin:
@nvonpentz either that, or i can move it under the transactions for it to sum as you had described.

maikelmclauflin:
yeah, pretty sure this can happen

evq:
perhaps we only change the `/accounts/balances` endpoint conditionally via a `pending=true` query parameter?  and forego changes to other APIs to keep things simple. I think in practice pending balances should be small so any discrepancy is unlikely to cause confusion. wdyt @maikelmclauflin / @nvonpentz

nvonpentz:
That sounds good to me.  Publishers will always supply `pending=true` for both displaying balances and creating payout reports.

maikelmclauflin:
yeah, that's fine

eyeshade/controllers/accounts.js
```diff
@@ -84,17 +109,30 @@ ORDER BY created_at
...
     if (!Array.isArray(accounts)) {
       accounts = [accounts]
     }
+    const args = [accounts]
 
-    const query1 = `select * from account_balances where account_id = any($1::text[])`
-
-    const transactions = await runtime.postgres.query(query1, [ accounts ])
-    reply(transactions.rows)
+    const votesPromise = pending ? runtime.postgres.query(selectPendingAccountVotes, args) : {
```
evq: can we short circuit this for non-channel accounts?

maikelmclauflin:
seems like travis is freezing up will try again in a couple of hours

PR #516, opened 2019-01-31
Supply client with surveyorIds from cohorts proportional to fund sources

### LINKED ISSUE #515: Precisely tag votes from transactions using mixed funds
If a user makes a contribution with grant and non grant funds, we tag all the votes that result from the transaction as grant when they should be tagged a mix of grant and control.

This happens because if there are any grant funds in the transaction, we only give the client surveyorIds from from grant surveyors.

https://github.com/brave-intl/bat-ledger/blob/9618b4fab8ce291443213c52e7e1ce3b3018318e/ledger/controllers/wallet.js#L349-L362

Instead we should be telling them to use grant and non grant surveyors proportional to the amount of funds in the transaction that were grant and non grant respectively.

nvonpentz:
Resolves #515  
Resolves https://github.com/brave-intl/bat-ledger/issues/65 (for now)

Assign surveyorsIds from grant and control cohorts proportional to fund sources in `PUT /v2/wallet/{paymentId}`

Also, remove legacy bitcoin wallet support (aka v1 wallet endpoints).

Lastly, clean up the end to end tests a fair amount, though not completely.  The e2e test file was effectively one large test, with many tests depending on state being altered by other tests.  This made it difficult to test new functionality (like contributions with mixed grant and user funds).  This pr:
* Separates each test so it can be run in isolation
* Abstracts common functionality like creating a wallet into their own functions
* Adds a test for contributions with mixed funds
* Forces the databases to be reset before every test

There are couple pieces of shared logic yet to be abstracted into a function and integrated into the tests:
* vote submission
* settle to publishers

### DIFF
--- a/ledger/controllers/wallet.js
+++ b/ledger/controllers/wallet.js
@@ -272,7 +272,7 @@ const write = function (runtime, apiVersion) {
     const viewings = runtime.database.get('viewings', debug)
     const wallets = runtime.database.get('wallets', debug)
 
-    let now, params, result, state, surveyor, surveyorIds, wallet, txnProbi, grantsProbiBeforeTx
+    let now, params, result, state, surveyor, surveyorIds, wallet, txnProbi
     let totalFee, grantFee, nonGrantFee
     let totalVotes, grantVotes, nonGrantVotes
 
@@ -336,9 +336,6 @@ const write = function (runtime, apiVersion) {
       }
     }
 
-    let { grants } = wallet
-    grantsProbiBeforeTx = grants ? (await sumActiveGrants(runtime, null, wallet, grants))[0] : new BigNumber(0)
-
     try {
       result = await runtime.wallet.redeem(wallet, txn, signedTx, request)
     } catch (err) {
@@ -362,14 +359,15 @@ const write = function (runtime, apiVersion) {
     if (result.status !== 'accepted' && result.status !== 'pending' && result.status !== 'completed') return reply(boom.badData(result.status))
 
     const grantIds = result.grantIds
+    const grantTotal = result.grantTotal
+
     if (grantIds) {
       await markGrantsAsRedeemed(grantIds)
-      let grantsProbiAfterTx = (await sumActiveGrants(runtime, null, wallet, grants))[0]
-      let grantsProbiUsedInTx = grantsProbiBeforeTx - grantsProbiAfterTx
-      grantVotes = new BigNumber(grantsProbiUsedInTx).dividedBy(params.probi).times(params.votes).round().toNumber()
+
+      grantVotes = new BigNumber(grantTotal).dividedBy(params.probi).times(params.votes).round().toNumber()
       nonGrantVotes = totalVotes - grantVotes
 
-      let grantProbiRate = grantsProbiUsedInTx / txnProbi
+      let grantProbiRate = grantTotal / txnProbi
       grantFee = totalFee * grantProbiRate
       nonGrantFee = totalFee - grantFee
 
--- a/ledger/controllers/wallet.js
+++ b/ledger/controllers/wallet.js
@@ -298,6 +298,10 @@ const write = function (runtime, apiVersion) {
       return reply(boom.badData(ex.toString()))
     }
 
+    params = surveyor.payload.adFree
+    txnProbi = runtime.wallet.getTxProbi(wallet, txn)
+    totalVotes = txnProbi.dividedBy(params.probi).times(params.votes).round().toNumber()
+
     if (!surveyor.cohorts) {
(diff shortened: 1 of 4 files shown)


ledger/controllers/wallet.js
```diff
@@ -290,7 +293,7 @@ const write = function (runtime, apiVersion) {
         return reply(boom.resourceGone('cannot perform a contribution using a legacy surveyor'))
       } else {
         // new contribution surveyor not yet populated with voting surveyors
-        const errMsg = 'surveyor ' + surveyor.surveyorId + ' has 0 surveyors, but needed ' + votes
```
nvonpentz: `votes` here will always be undefined.  do we want to calculate votes earlier then we do currently?

ledger/controllers/wallet.js
```diff
@@ -323,6 +326,9 @@ const write = function (runtime, apiVersion) {
       }
     }
 
+    let { grants } = wallet
```
nvonpentz: not sure how this syntax works seeing as grants is not an attribute on wallet documents

nvonpentz: as discussed on slack, `grants` is also a field on the wallet documents.  when claimed, a grant is moved to the wallet document.

ledger/controllers/wallet.js
nvonpentz: I defined `totalVotes` earlier than this block so it should always be defined now https://github.com/brave-intl/bat-ledger/pull/516/commits/6d4a5bdb0f29d9c372298aa65eb50be7cba2aaa9

ledger/controllers/wallet.js
```diff
@@ -295,12 +298,16 @@ const write = function (runtime, apiVersion) {
       return reply(boom.badData(ex.toString()))
     }
 
+    params = surveyor.payload.adFree
+    txnProbi = runtime.wallet.getTxProbi(wallet, txn)
+    totalVotes = txnProbi.dividedBy(params.probi).times(params.votes).round().toNumber()
+
```
evq: can we add a sanity check here to ensure `totalVotes >= 1` and error if not?

ledger/controllers/wallet.js
```diff
@@ -309,22 +316,18 @@ const write = function (runtime, apiVersion) {
       }
     }
 
-    params = surveyor.payload.adFree
-
-    votes = runtime.wallet.getTxProbi(wallet, txn).dividedBy(params.probi).times(params.votes).round().toNumber()
-
-    if (votes < 1) votes = 1
+    if (totalVotes < 1) totalVotes = 1
```
evq: we can remove this if we add the check described above

test/e2e.integration.test.js
```diff
@@ -360,6 +360,24 @@ test('ledger : v2 contribution workflow with uphold BAT wallet', async t => {
...
   t.true(surveyorIds.length >= 5)
 
+ // look up surveyorIds to ensure that they belong to the correct cohorts
+  const ledgerDB = await connectToDb('ledger')
+  const surveyors = ledgerDB.collection('surveyors')
+
+  let numControlSurveryors = 0
+  let numGrantSurveyors = 0
+  for (let surveyorId of surveyorIds) {
+    let cohort = (await surveyors.findOne({surveyorId: surveyorId})).payload.cohort
```
maikelmclauflin: Nit: No need to keep it all on one line. Would appreciate doing just one step at a time.
1) do the query
2) access the cohort value

ledger/controllers/wallet.js
nvonpentz: https://github.com/brave-intl/bat-ledger/pull/516/commits/5324e8c0dd75d8fb6145ffadc68ba0874e077d63

ledger/controllers/wallet.js
```diff
@@ -379,28 +368,39 @@ const write = function (runtime, apiVersion) {
...
+
+    if (grantVotes > 0) {
+      await runtime.queue.send(debug, 'contribution-report', underscore.extend({
+        paymentId: paymentId,
+        address: wallet.addresses[result.altcurrency],
+        surveyorId: surveyorId,
+        viewingId: viewingId,
+        fee: grantFee,
+        votes: grantVotes,
+        cohort: 'grant'
```
evq: I think this should be grantCohort

evq:
one nit, otherwise LGTM pending the additional mixed funds tests you mentioned

ledger/controllers/wallet.js
nvonpentz: resolved in https://github.com/brave-intl/bat-ledger/pull/516/commits/f062d93b5091fe00ab2a04494ac47d7fccd32b44

PR #517, opened 2019-01-31
updated test to fix random failure

maikelmclauflin:


package.json
```diff
@@ -32,6 +32,7 @@
     "test-integration": "ava -v -s test/*/*.integration.test.js && ava -v test/e2e.integration.test.js",
     "test-unit": "ava -v",
     "docker-reset": "docker-compose down && docker rmi $(docker images -a -q) --force",
+    "docker-down": "npm run db-reset && docker-compose down",
```
nvonpentz: Are we sure this will actually reset the databases used in docker? Looks like `npm run db-reset` will call `npm run mongo-reset && npm run redis-flush`

nvonpentz: We don't have any volumes for mongo or redis so I _think_ `docker-compose down` should be sufficient to reset database

maikelmclauflin: sorry, don't we have `ledger-redis`?

maikelmclauflin: ah. gotcha

PR #520, opened 2019-02-05
[WIP] Expose endpoint to publishers for inserting manual transactions

### LINKED ISSUE #518: expose endpoint to create manual transactions from publishers app
will be used to track manual publisher payments as part of - https://github.com/brave-intl/publishers/issues/1571

nvonpentz:
Resolves #518 

Expose an endpoint
```
PUT /v1/accounts/{account_type}/{account_id}/transactions/{transaction_type}/{transaction_id}
```
to publishers that inserts a transaction from the settlement wallet address to the owner identifier.

All transactions will have type `manual`, be from the settlement wallet, and to an owner account that publishers supplies.

At the moment we only accept `account_types` that are `owners`, and transaction types that are `manual`, so we might not want to generalize those as variables in the endpoint uri.

### DIFF
--- a/eyeshade/controllers/accounts.js
+++ b/eyeshade/controllers/accounts.js
@@ -376,6 +376,59 @@ v1.getPaidTotals =
 }
 
 /*
+  PUT /v1/accounts/{account_type}/{account_id}/transactions/{transaction_type}/{transaction_id}
+*/
+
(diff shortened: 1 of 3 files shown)


nvonpentz:
See #518

PR #524, opened 2019-02-07
added bee-queue to replace pho-queue

### LINKED ISSUE #522: convert rsmq to qless
WHY
logic for queue setup should exist in redis using a package like `qless-js`

WHERE
install the package as a dependency and utilize it in the worker files and other files such as the `runtime-queue` file. the pattern laid out in the `readme.md` of `qless-js` should be sufficient

maikelmclauflin:
implemented bee-queue as qless did not have a robust worker implementation / documentation fixes #522

### DIFF
--- a/bat-utils/lib/runtime-queue.js
+++ b/bat-utils/lib/runtime-queue.js
@@ -1,181 +1,149 @@
-const bluebird = require('bluebird')
+const BeeQueue = require('bee-queue')
 const redis = require('redis')
-const Rsmq = require('rsmq')
-const RsmqWorker = require('rsmq-worker')
 const SDebug = require('sdebug')
 const debug = new SDebug('queue')
-const underscore = require('underscore')
-
-bluebird.promisifyAll(redis.RedisClient.prototype)
-bluebird.promisifyAll(redis.Multi.prototype)
-
-const Queue = function (config, runtime) {
-  if (!(this instanceof Queue)) return new Queue(config, runtime)
-
-  if (!config.queue) return
-
-  this.config = config.queue.rsmq || config.queue
-  if (typeof this.config === 'string') {
-    if (this.config.indexOf('redis://') === -1) this.config = 'redis://' + this.config
-    this.config = { client: redis.createClient(this.config) }
-  }
-  this.rsmq = new Rsmq(this.config)
-  this.runtime = runtime
-
-  this.rsmq.on('connect', () => { debug('redis connect') })
-    .on('disconnect', () => { debug('redis disconnect') })
-    .on('error', (err) => {
-      debug('redis error', err)
-      this.runtime.captureException(err)
-    })
-}
-
-Queue.prototype.create = async function (name) {
-  const self = this
-
-  return new Promise((resolve, reject) => {
-    self.rsmq.listQueues((err, rsp) => {
-      if (err) {
-        debug('listQueues failed')
-        return reject(err)
-      }
-      if (rsp.indexOf(name) !== -1) return resolve(false)
-
-      self.rsmq.createQueue({ qname: name }, (err, rsp) => {
-        if (err) {
-          debug('createQueue ' + name + ' failed')
-          return reject(err)
-        }
-
-        if (rsp !== 1) return reject(new Error('createQueue ' + name + ' failed: unknown response'))
-        resolve(true)
-      })
-    })
-  })
-}
-
-Queue.prototype.drop = async function (name) {
-  const self = this
-
-  return new Promise((resolve, reject) => {
-    self.rsmq.listQueues((err, rsp) => {
-      if (err) {
-        debug('listQueues failed')
-        return reject(err)
-      }
-      if (rsp.indexOf(name) === -1) return resolve(false)
-
-      self.rsmq.deleteQueue({ qname: name }, (err, rsp) => {
-        if (err) {
-          debug('deleteQueue ' + name + ' failed')
-          return reject(err)
-        }
-
-        if (rsp !== 1) return reject(new Error('deleteQueue ' + name + ' failed: unknown response'))
-        resolve(true)
-      })
-    })
-  })
-}
-
-Queue.prototype.send = async function (debug, name, payload) {
-  const self = this
-
-  return new Promise((resolve, reject) => {
-    self.rsmq.sendMessage({ qname: name, message: JSON.stringify(payload) }, (err, rsp) => {
-      if (err) {
-        debug('sendMessage ' + name + ' failed', payload)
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -64,6 +64,7 @@
     "accept-language-parser": "1.4.1",
     "bat-publisher": "^2.0.24",
     "bat-utils": "file:./bat-utils",
+    "bee-queue": "^1.2.2",
```
nvonpentz: Would be good to security team's approval here too

eyeshade/workers/surveyors.js
```diff
@@ -44,7 +45,7 @@ exports.workers = {
         `
         const votingQ = await client.query(query1, [surveyorId])
         if (!votingQ.rowCount) {
-          throw new Error('no votes for this surveyor!')
+          return runtime.captureException(new Error('no votes for this surveyor!'))
```
nvonpentz: When do we need to explicitly wrap errors with captureException? My understanding was that sentry would capture these automatically

bat-utils/lib/runtime-queue.js
```diff
@@ -1,181 +1,80 @@
...
-  })
+function Queue (config, runtime) {
+  const { queue: url } = config
+  this.runtime = runtime
+  this.created = {}
+  this.config = {
+    url,
+    client: redis.createClient(url)
+  }
+  console.log('url for queue', url)
```
nvonpentz: should we keep this log statement?

eyeshade/workers/surveyors.js
```diff
@@ -1,6 +1,7 @@
 const { insertFromVoting, updateBalances } = require('../lib/transaction.js')
 const { mixer } = require('../workers/reports.js')
 
+exports.name = 'surveyors'
```
nvonpentz: I noticed we have similar lines in many worker files, though I don't know what this does.  Could you explain what these lines are doing?

nvonpentz:
I know one of the original goals for the rsmq replacement was to have a web interface.  Have you considered integrating https://github.com/bee-queue/arena?

bat-utils/lib/runtime-queue.js
```diff
@@ -1,181 +1,80 @@
...
 }
 
-Queue.prototype.drop = async function (name) {
-  const self = this
-
+async function send (debug, key, data) {
+  debug(`sending ${key}: ${JSON.stringify(data)}`)
+  const client = this.created[key] || this.create(key)
+  const job = await client.createJob(data).save()
```
nvonpentz: Do we want to specify .retries() here?

bat-utils/lib/runtime-queue.js
```diff
@@ -1,181 +1,80 @@
...
-        return reject(err)
-      }
-      if (rsp.indexOf(name) !== -1) return resolve(false)
-
-      self.rsmq.createQueue({ qname: name }, (err, rsp) => {
-        if (err) {
-          debug('createQueue ' + name + ' failed')
-          return reject(err)
-        }
+module.exports = Queue
```
nvonpentz: I'm noticing that module.exports is defined in different places in different files throughout the codebase.  Ideally we can keep this consistent (e.g. always at the bottom of the file).  Where is the best place to have this line?

maikelmclauflin: So, i like to let all function declarations be hoisted up by js so that you can see a summary of the file's io quickly without having to look through the minutia of what the fns are actually doing. I believe marshal had previously set up the files to be definitions where no variable was available until it was declared as such. We can talk about it which to use and when in group dms

bat-utils/lib/runtime-queue.js
maikelmclauflin: it can be removed

bat-utils/lib/runtime-queue.js
maikelmclauflin: yes, that does make sense to do

eyeshade/workers/surveyors.js
maikelmclauflin: this is a legacy line from the previous implementations that marshal had setup. they can be removed. i was just trying to be consistent until we do so

eyeshade/workers/surveyors.js
maikelmclauflin: this was when i was debugging. thanks

package.json
maikelmclauflin: yes

eyeshade/workers/surveyors.js
maikelmclauflin: oh, no. this was because the initializer fn that triggers this waits for it to succeed.

maikelmclauflin:
> I know one of the original goals for the rsmq replacement was to have a web interface. Have you considered integrating https://github.com/bee-queue/arena?

yes @nvonpentz . have not gotten to it yet though

maikelmclauflin:
https://github.com/brave/internal/issues/524#event-2169166234

bat-utils/lib/runtime-queue.js
nvonpentz: I'm cool whatever you think is best, just hoping to be consistent

eyeshade/workers/surveyors.js
nvonpentz: not sure I entirely understand, do you mind explaining what would happen if we threw here instead of returning early and using captureException?

maikelmclauflin: so, here for instance: https://github.com/brave-intl/bat-ledger/blob/master/eyeshade/workers/reports.js#L52 if we have 2 surveyors and the first doesn't have any votes (nothing to do) it will throw instead of just completing the actions for zero items, which causes our second surveyor not to go through the process

nvonpentz:
From the bee-queue [doc](https://github.com/bee-queue/bee-queue#installation):
> * We've been noticing that some jobs get delayed by virtue of an issue with Redis < 3.2, and therefore recommend the use of Redis 3.2+.

It looks like we're on redis 2.8, should we upgrade?

maikelmclauflin:
i think we're on `3.2.12` according to the datastore's version

nvonpentz:
can we remove rsmq and rsmq-worker from package json?

nvonpentz:
spoke with michael, my approval was premature - still need to remove some deps and retry logic

maikelmclauflin:
Sorry, a couple more things needed before this can go live.

request from @evq 

```
verify that the bee-queue implementation does not drop jobs if the worker process is killed instead of reports an error
```

tsmartt:
Closing as stale.

PR #528, opened 2019-02-11
added alt account_types

### LINKED ISSUE #525: add account types for non BAT currencies
`ethereum`, `litecoin`, and `bitcoin` - `account_types` should be added and integrated into server logic.
as outlined here: https://github.com/brave-intl/internal/issues/77

### LINKED ISSUE #526: create user deposit transaction type
`user_deposit` - `transaction_type` should be created on `transaction` table
as outlined here: https://github.com/brave-intl/internal/issues/77

maikelmclauflin:
fixed down migrations script and added account types to schema and added account_types
fixes #525 
fixes #526

### DIFF
--- a/eyeshade/controllers/accounts.js
+++ b/eyeshade/controllers/accounts.js
@@ -14,6 +14,11 @@ const braveJoi = utils.extras.joi
 
 const v1 = {}
 
+const settlementDestinationTypes = ['uphold']
+const accountTypes = ['channel', 'owner'].concat(settlementDestinationTypes)
+const transactionTypes = ['contribution', 'referral', 'contribution_settlement', 'referral_settlement', 'fees', 'scaleup', 'manual', 'user_deposit', 'manual_payout']
+
+const accountTypeValidation = Joi.string().valid(accountTypes)
 const orderParam = Joi.string().valid('asc', 'desc').optional().default('desc').description('order')
 const joiChannel = Joi.string().description('The channel that earned or paid the transaction')
 const joiPaid = Joi.number().description('amount paid out in BAT')
@@ -96,9 +101,9 @@ ORDER BY created_at
       amount: Joi.number().required().description('amount in BAT'),
       settlement_currency: braveJoi.string().anycurrencyCode().optional().description('the fiat of the settlement'),
       settlement_amount: Joi.number().optional().description('amount in settlement_currency'),
-      settlement_destination_type: Joi.string().optional().valid(['uphold']).description('type of address settlement was paid to'),
+      settlement_destination_type: Joi.string().optional().valid(settlementDestinationTypes).description('type of address settlement was paid to'),
       settlement_destination: Joi.string().optional().description('destination address of the settlement'),
-      transaction_type: Joi.string().valid(['contribution', 'referral', 'contribution_settlement', 'referral_settlement', 'fees', 'scaleup', 'manual']).required().description('type of the transaction')
+      transaction_type: Joi.string().valid(transactionTypes).required().description('type of the transaction')
     }))
   }
 }
@@ -107,7 +112,6 @@ ORDER BY created_at
    GET /v1/accounts/balances/{type}/top
 */
 
-const accountTypes = ['channel', 'owner', 'uphold']
 v1.getTopBalances =
 { handler: (runtime) => {
   return async (request, reply) => {
@@ -137,7 +141,7 @@ v1.getTopBalances =
 
   validate: {
     params: Joi.object().keys({
-      type: Joi.string().required().valid(accountTypes).description('balance types to retrieve')
+      type: accountTypeValidation.required().description('balance types to retrieve')
     }),
     query: {
       limit: Joi.number().min(1).default(10).description('the top balances to retrieve')
@@ -148,7 +152,7 @@ v1.getTopBalances =
     schema: Joi.array().items(
       Joi.object().keys({
         account_id: Joi.string(),
-        account_type: Joi.string().valid(accountTypes),
+        account_type: accountTypeValidation,
         balance: Joi.number().description('balance in BAT')
       })
     )
@@ -220,7 +224,7 @@ v1.getBalances = {
     schema: Joi.array().items(
        Joi.object().keys({
          account_id: Joi.string(),
-         account_type: Joi.string().valid(['channel', 'owner', 'uphold']),
+         account_type: Joi.string().valid(accountTypes),
          balance: Joi.number().description('balance in BAT')
        })
      )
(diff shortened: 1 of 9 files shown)


eyeshade/controllers/accounts.js
```diff
@@ -96,9 +101,9 @@ ORDER BY created_at
       amount: Joi.number().required().description('amount in BAT'),
       settlement_currency: braveJoi.string().anycurrencyCode().optional().description('the fiat of the settlement'),
       settlement_amount: Joi.number().optional().description('amount in settlement_currency'),
-      settlement_destination_type: Joi.string().optional().valid(['uphold']).description('type of address settlement was paid to'),
+      settlement_destination_type: Joi.string().optional().valid(settlementDestinationTypes).description('type of address settlement was paid to'),
```
evq: I'm a bit confused about the intended use of this API given these changes.

this was originally intended to be scoped to settlement type transactions, adding the various on-chain accounts as potential settlement destination types currently doesn't make sense

also using this to display user deposits doesn't seem to fit since the destination will always be the user card, not an on-chain address

eyeshade/migrations/0005_account_types/down.sql
```diff
@@ -0,0 +1,95 @@
...
+
+alter table transactions add constraint check_from_account_type
+  check (from_account_type in ('channel', 'owner', 'uphold', 'internal', 'payment_id'));
+alter table transactions add constraint check_to_account_type
+  check (to_account_type in ('channel', 'owner', 'uphold', 'internal', 'payment_id'));
+
+alter table transactions add constraint check_transaction_type
+  check (transaction_type in ('contribution', 'referral', 'contribution_settlement', 'referral_settlement', 'fees', 'scaleup', 'manual', 'ad', 'ad_settlement', 'manual_settlement'));
+
+create view account_transactions(
```
evq: I'm confused about this - we didn't delete the view in `up.sql`. also shouldn't this file have a guard per your other recent changes to the `down.sql` files?

maikelmclauflin: yes. looks like i didn't push that change.

eyeshade/controllers/accounts.js
maikelmclauflin: ah. gotcha. looks like i was getting ahead of myself then. so just the transaction types below then

eyeshade/lib/transaction.js
```diff
@@ -1,27 +1,98 @@
 const BigNumber = require('bignumber.js')
 const getPublisherProps = require('bat-publisher').getPublisherProps
 const uuidv5 = require('uuid/v5')
+const uuid = require('uuid')
```
evq: can you change this in light of the notice on https://www.npmjs.com/package/uuid
> [Deprecation warning: The use of require('uuid') is deprecated and will not be supported after version 3.x of this module. Instead, use require('uuid/[v1|v3|v4|v5]') as shown in the examples below.]

eyeshade/lib/transaction.js
```diff
@@ -1,27 +1,98 @@
...
+  amount = new BigNumber(amount)
+  if (amount.lessThanOrEqualTo(0)) {
+    return [] // skip because we don't track tx's that don't have an accounting purpose
+  }
+
+  const {
+    rows
+  } = await client.query(insertTransaction, [
+    // surveyorId and channel pair should be unique
+    uuid.v4(),
```
evq: what do you think about using uuid.v5 based on the on chain hash?

eyeshade/lib/transaction.js
```diff
@@ -1,27 +1,98 @@
...
+    createdAt / 1000,
+    `deposits from ${chain} chain`,
+    'user_deposit',
+    id,
+    address,
+    chain,
+    owner,
```
evq: using `owner` in this context is confusing, it conflicts with our use of `owner` to refer to a publisher's app account

eyeshade/lib/transaction.js
```diff
@@ -1,27 +1,98 @@
...
 const uuidv5 = require('uuid/v5')
+const uuid = require('uuid')
 const {
   createdTimestamp,
   normalizeChannel
 } = require('bat-utils/lib/extras-utils')
-
 const getYoutubeChannelId = require('bat-utils/lib/youtube')
 
+const knownChains = {
```
evq: currently this doesn't allow for differentiating BAT deposited from an ETH addr and ETH deposited that was converted to BAT

eyeshade/lib/transaction.js
```diff
@@ -1,27 +1,98 @@
...
+  created_at,
+  description,
+  transaction_type,
+  document_id,
+  from_account,
+  from_account_type,
+  to_account,
+  to_account_type,
+  amount,
+  settlement_currency,
```
evq: I would prefer we stick with the current pattern of using queries that don't include columns we're going to set to `null` or use the default value. I think it's clearer when the query is embedded in the context as well.

eyeshade/lib/transaction.js
evq: ah I see the test is actually making a deposit from on-chain to an owner

to clarify a bit we're interested in tracking user deposits into the anonymous uphold cards - so perhaps `cardId` would fit better. also maybe we should rename this function to `insertUserDepositFromChain` so it's more clear that this isn't for general on-chain transactions - just user deposits

eyeshade/lib/transaction.js
maikelmclauflin: mmm. makes sense. can do. i'll have to make sure that that is not happening in other prs too

eyeshade/lib/transaction.js
maikelmclauflin: yeah, sure. that makes sense. so that we can't have 2 of the same tx ids

eyeshade/lib/transaction.js
maikelmclauflin: makes sense. can do

eyeshade/lib/transaction.js
maikelmclauflin: does that make sense? we will have 8 queries in this one file that do almost exactly the same thing. i would think that repeating similar queries takes away from readability. would putting the insertion step into a function be preferable?

eyeshade/lib/transaction.js
maikelmclauflin: do we have information for this? sorry, just not sure what / where we will be getting info from

maikelmclauflin: i suppose we would be hitting: https://uphold.com/en/developer/api/documentation/#list-user-transactions

PR #530, opened 2019-02-12
used timestamp for created at value

### LINKED ISSUE #529: add created_at for settlement endpoint
`POST /v2/publishers/settlement` should be able to receive a `created at` value to allow for manual settlements

maikelmclauflin:
fixes #529

### DIFF
--- a/eyeshade/controllers/publishers.js
+++ b/eyeshade/controllers/publishers.js
@@ -23,26 +23,36 @@ v2.settlement = {
       const debug = braveHapi.debug(module, request)
       const settlements = runtime.database.get('settlements', debug)
       const fields = [ 'probi', 'amount', 'fee', 'fees', 'commission' ]
-      let entry, state
+      let transactionId = null
 
-      for (entry of payload) {
+      for (let entry of payload) {
         if (entry.altcurrency !== altcurrency) return reply(boom.badData('altcurrency should be ' + altcurrency))
       }
 
-      state = {
-        $currentDate: { timestamp: { $type: 'timestamp' } },
-        $set: {}
-      }
-      for (entry of payload) {
+      const $currentDate = { timestamp: { $type: 'timestamp' } }
+      const executedTime = new Date()
+      for (let entry of payload) {
+        const {
+          executedAt,
+          transactionId: settlementId
+        } = entry
+        transactionId = settlementId
+        const $set = {}
+        const state = {
+          $currentDate,
+          $set
+        }
         entry.commission = new BigNumber(entry.commission).plus(new BigNumber(entry.fee)).toString()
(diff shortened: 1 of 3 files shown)


eyeshade/controllers/publishers.js
```diff
@@ -59,6 +65,7 @@ v2.settlement = {
 
   validate: {
     payload: Joi.array().min(1).items(Joi.object().keys({
+      timestamp: braveJoi.date().iso().optional().description('the timestamp the settlement occurred'),
```
evq: what do you think about calling this `executed_at`?

maikelmclauflin: sure. yeah, happy to do any key

maikelmclauflin: @evq did you just want the input key to change or should it be stored on the document in that way too

evq: I think changing the field stored on the document would be good too :+1:

maikelmclauflin: would you be ok with `executedAt`? just to keep camel case consistent

evq: ah good point! sgtm :+1:

maikelmclauflin:
ya. can do

test/e2e.integration.test.js
```diff
@@ -236,6 +236,28 @@ test('ledger : user contribution workflow with uphold BAT wallet', async t => {
...
+FROM transactions
+WHERE
+    transaction_type = 'contribution_settlement';
+`
+  const {
+    rows
+  } = await postgres.query(select)
+
+  t.deepEqual(rows.map((entry) => _.omit(entry, ['from_account', 'to_account', 'document_id', 'id'])), [{
+    created_at: new Date(+newYear + 2000),
```
evq: looks like `newYear` isn't defined anywhere?

maikelmclauflin: yeah. my bad. not sure how that got through the commit. just uploaded

eyeshade/controllers/publishers.js
```diff
@@ -149,6 +159,7 @@ module.exports.initialize = async (debug, runtime) => {
         commission: bson.Decimal128.POSITIVE_ZERO,    // conversion fee (i.e., for settlement)
         fee: bson.Decimal128.POSITIVE_ZERO,           // network fee (i.e., for settlement)
         type: '',
+        executedAt: bson.Timestamp.ZERO,              // When the settlement was executed
```
evq: should this be `bson.Date.ZERO`?

maikelmclauflin: Date input will be turned into and stored as timestamp by mongo

PR #532, opened 2019-02-15
Fix migration 0004

nvonpentz:
Add missing sql insert into migrations table.

Deploy plan:

1. Run 0004 down
2. Merge this PR and auto deploy to staging
3. Run 0004 up

### DIFF
--- a/eyeshade/migrations/0004_manual_settlements/up.sql
+++ b/eyeshade/migrations/0004_manual_settlements/up.sql
@@ -1,4 +1,8 @@
+select execute($$
+
 insert into migrations (id, description) values ('0004', 'manual_settlements');
 alter table transactions drop constraint check_transaction_type;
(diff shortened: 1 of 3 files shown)


nvonpentz:
I'm going to see if I can write a test to ensure everything matches from now on

eyeshade/migrations/0004_manual_settlements/up.sql
```diff
@@ -1,3 +1,4 @@
+insert into migrations (id, description) values ('0004', 'manual_settlements');
```
maikelmclauflin: can we surround this with the logic present in other migrations?

maikelmclauflin: i'll write up a new pr to do the same for the down files

PR #534, opened 2019-02-15
added down logic for brutish behavior

maikelmclauflin:


eyeshade/migrations/0000_initial/down.sql
```diff
@@ -1 +1,5 @@
+select execute($$
+
 drop table migrations;
+
+$$) where exists (select * from migrations where id = '0001');
```
evq: should this be 0000?

eyeshade/migrations/0003_ads/down.sql
```diff
@@ -90,3 +92,5 @@ create materialized view account_balances(
 create unique index on account_balances(account_type, account_id);
 
 refresh materialized view account_balances;
+
+$$) where exists (select * from migrations where id = '0004');
```
evq: should this be 0003?

eyeshade/migrations/0000_initial/down.sql
maikelmclauflin: yis

eyeshade/migrations/0003_ads/down.sql
maikelmclauflin: yis

PR #535, opened 2019-02-15
Replace youtube username with channel id at transaction insertion time

nvonpentz:
Resolves https://github.com/brave-intl/internal/issues/79

Running the surveyor-frozen-report should insert all the mislabeled transactions with the correct identifiers.

This adds a test that makes an external request to the youtube data api.  Please let me know if that's not desired.


bat-utils/lib/youtube.js
```diff
@@ -0,0 +1,15 @@
+'use strict'
+
+const braveHapi = require('./extras-hapi')
+
+async function getYoutubeChannelId (youtubeUsername) {
+  const youtubeApiKey = process.env.YOUTUBE_API_KEY
+  const youtubeApiBaseUri = process.env.YOUTUBE_API_BASE_URI || 'https://www.googleapis.com'
```
nvonpentz: Should this be an envar? It is on publishers but I don't really see this changing.

bat-utils/lib/youtube.js
```diff
@@ -0,0 +1,15 @@
...
+
+  let response = await braveHapi.wreck.get(`${youtubeApiBaseUri}/youtube/v3/channels?part=id&forUsername=${youtubeUsername}&key=${youtubeApiKey}`)
+  response = JSON.parse(response)
+
+  return response.items[0].id
+}
+
+module.exports = getYoutubeChannelId
```
nvonpentz: Should this be it's own file or just be a function extras-utils?

bat-utils/lib/youtube.js
maikelmclauflin: might as well keep it as such. quicker update times

bat-utils/lib/youtube.js
maikelmclauflin: i'm fine with either, even going down to one function per file

test/eyeshade/surveyor.integration.test.js
```diff
@@ -44,50 +44,66 @@ test('verify frozen occurs when daily is run', async t => {
...
+})
 
-  async function tryFreeze (dayShift, expect) {
-    await freezeOldSurveyors(debug, runtime, dayShift)
-    // beware of cursor
-    const surveyor = await querySurveyor()
-    t.is(surveyor.rowCount, 1)
-    t.is(surveyor.rows[0].frozen, expect)
-  }
+test.only('replaces invalid youtube user ids with youtube channel ids when inserting into transaction table', async t => {
```
maikelmclauflin: don't forget to remove the only

nvonpentz: ah thank you!

bat-utils/lib/youtube.js
nvonpentz: radical

PR #536, opened 2019-02-15
added tests for non ad grant list

maikelmclauflin:
modified grant logic to only return promos that are actually available to user when filtered through their providerId


ledger/controllers/grants.js
```diff
@@ -1114,24 +1114,51 @@ async function chooseFromAvailablePromotions (debug, runtime, {
...
+  const promoCheckers = [
+    getAdsPromos,
+    getUgpPromos
+  ]
+  for (let available of promoCheckers) {
+    const promos = await available(debug, runtime, options)
+    promotions = promotions.concat(promos)
+  }
+  return promotions
+}
```
nvonpentz: have you considered writing this function without the loop?

maikelmclauflin: yeah, that's how i started it. it just seemed redundant and made more sense (to me) to produce a loop to add more types as needed

PR #541, opened 2019-02-22
added coverage

maikelmclauflin:


package.json
```diff
@@ -120,6 +122,8 @@
   "devDependencies": {
     "ava": "^1.0.0-beta.0",
     "babel-eslint": "^8.2.3",
+    "coveralls": "^3.0.3",
+    "nyc": "^13.3.0",
```
nvonpentz: We probably need to get a quick +1 from security for the new deps

package.json
```diff
@@ -21,7 +21,9 @@
     "start-eyeshade-worker": "NODE_PATH=. DEBUG=*,-babel*,-mongo:*,mongo:queries SERVICE=eyeshade ./node_modules/.bin/babel-node --optimize_for_size --max_old_space_size=460 --gc_interval=100 eyeshade/worker.js",
     "start-ledger": "NODE_PATH=. DEBUG=*,-babel*,-mongo:*,mongo:queries SERVICE=ledger ./node_modules/.bin/babel-node ledger/server.js",
     "start-ledger-worker": "NODE_PATH=. DEBUG=*,-babel*,-mongo:*,mongo:queries SERVICE=ledger ./node_modules/.bin/babel-node --optimize_for_size --max_old_space_size=460 --gc_interval=100 ledger/worker.js",
-    "test": "npm run test-unit && npm run test-integration",
+    "test": "nyc --reporter=html --reporter=text npm run test-all",
+    "coverage": "nyc report --reporter=text-lcov | coveralls",
+    "test-all": "npm run test-unit && npm run test-integration",
```
nvonpentz: Do we support development / testing that isn't using docker?

maikelmclauflin: yeah, so you can run tests outside and point your env vars to the correct values (localhost) like what happens in travis which will hit the docker containers https://github.com/brave-intl/bat-ledger/blob/master/.travis.yml#L10 even if the tests are not running inside of a container. there is certainly an argument for consistent testing locally vs on travis as it makes the model of what is happening less complex, but we can fix that in another pr.

package.json
maikelmclauflin: true

maikelmclauflin: https://github.com/brave/internal/issues/520

maikelmclauflin:
review has completed: https://github.com/brave/internal/issues/520#issuecomment-467575178

PR #543, opened 2019-02-26
added default 0 to balance value accessing

### LINKED ISSUE #465: use default 0 for undefined balance
https://sentry.io/brave-software/ledger-production/issues/615457752/?query=is%3Aunresolved

maikelmclauflin:
fixes #465

### DIFF
--- a/bat-utils/lib/runtime-wallet.js
+++ b/bat-utils/lib/runtime-wallet.js
@@ -345,8 +345,9 @@ Wallet.providers.uphold = {
       throw ex
     }
 
-    const balanceProbi = new BigNumber(cardInfo.balance).times(this.currency.alt2scale(info.altcurrency))
-    const spendableProbi = new BigNumber(cardInfo.available).times(this.currency.alt2scale(info.altcurrency))
+    const altScale = this.currency.alt2scale(info.altcurrency)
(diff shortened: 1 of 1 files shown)


bat-utils/lib/runtime-wallet.js
```diff
@@ -346,8 +351,18 @@ Wallet.providers.uphold = {
     }
 
     const altScale = this.currency.alt2scale(info.altcurrency)
-    const balanceProbi = new BigNumber(cardInfo.balance || 0).times(altScale)
-    const spendableProbi = new BigNumber(cardInfo.available || 0).times(altScale)
+    const { error } = Joi.validate(cardInfo, cardInfoSchema)
+    if (error) {
+      this.runtime.captureException(error, {
```
evq: would it be more useful to check this inside bat-utils where we have more context?

maikelmclauflin: Sorry, i don't understand what more context we would have. I'm happy to, I just don't understand the reason.

evq: we have the full response from uphold

maikelmclauflin: yeah, we have the response in this function. so that means we have more context, no?

evq: i'm blind :)

PR #555, opened 2019-03-05
moved publisher normalization

### LINKED ISSUE #554: resolve id constraint issues with youtube user lookup and re-enable
was temporarily disabled in https://github.com/brave-intl/bat-ledger/pull/553

maikelmclauflin:
work in progress for catching publisher syntax errors

fixes #554


nvonpentz:
I think this is the right place so that the votes are consistent throughout every step going forward.  We do have the those existing votes though that will need to be changed still.  We could do that manually or do a similar lookup at insertion time that modified the original vote records.

PR #576, opened 2019-03-14
Make PUT /v1/referrals/{transactionID} idempotent

### LINKED ISSUE #574: referral finalization endpoint is not idempotent
If promo sends finalized referrals that eyeshade already has, eyeshade returns a 422.  This locks up the referral server which will continually try to send the same download ids and revert once it fails.

Instead of returning a 422 when eyeshade sends downloads that have already been confirmed, let's accept them and return 200.

nvonpentz:
Also add tests for referrals
Resolves #574
Let me know if I left too many comments.

### DIFF
--- a/eyeshade/controllers/referrals.js
+++ b/eyeshade/controllers/referrals.js
@@ -77,9 +77,6 @@ v1.createReferrals = {
       const downloadIdsToBeConfirmed = []
       let entries, existingDownloadIds, probi
 
-      entries = await referrals.find({ transactionId: transactionId })
-      if (entries.length > 0) return reply({})
-
       probi = await runtime.currency.fiat2alt(runtime.config.referrals.currency, runtime.config.referrals.amount, altcurrency)
       probi = bson.Decimal128.fromString(probi.toString())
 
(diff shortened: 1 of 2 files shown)


eyeshade/controllers/referrals.js
```diff
@@ -74,29 +74,39 @@ v1.createReferrals = {
       const payload = request.payload
       const debug = braveHapi.debug(module, request)
       const referrals = runtime.database.get('referrals', debug)
-      let entries, matches, probi, query
+      const downloadIdsToBeConfirmed = []
+      let entries, existingDownloadIds, probi
 
       entries = await referrals.find({ transactionId: transactionId })
-      if (entries.length > 0) return reply(boom.badData('existing transaction-identifier: ' + transactionId))
+      if (entries.length > 0) return reply({})
```
evq: I wonder if we should avoid early return here. as it's possible (although unlikely) that some referrals were added to the collection and others were not

nvonpentz: Yes, you're right.  I don't think we care if there already is a referral with the same transaction id.  We should only care about download id.  I will remove this line.

In practice though I don't think this should be too much of a problem since txId is effectively a request id, since a new one is used every request.  So in the case you described, the referrals should just go through on the next request.

nvonpentz: Added a fix + test in 07c466d

PR #584, opened 2019-03-19
added test for owners

### LINKED ISSUE #565: owner wallet endpoint returns rates even for invalid owner ids

maikelmclauflin:
fixes #565

### DIFF
--- a/eyeshade/controllers/owners.js
+++ b/eyeshade/controllers/owners.js
@@ -25,12 +25,16 @@ v1.getWallet = {
       const owners = runtime.database.get('owners', debug)
       let entry, provider, rates, result
 
+      entry = await owners.findOne({ owner })
+      provider = entry && entry.provider
+      if (!provider) {
+        return reply(boom.notFound('owner does not exist'))
+      }
+
       result = {
         rates: await runtime.currency.rates(altcurrency)
       }
 
-      entry = await owners.findOne({ owner: owner })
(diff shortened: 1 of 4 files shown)


eyeshade/controllers/owners.js
```diff
@@ -51,6 +55,7 @@ v1.getWallet = {
       } catch (ex) {
         debug('status', { reason: ex.toString(), stack: ex.stack })
         runtime.captureException(ex, { req: request, extra: { owner: owner } })
+        return reply(boom.boomify(ex))
```
evq: I don't think we want to return error here

maikelmclauflin: oh, this is just so that the subsequent lines are not hit, so that we don't get the reply called twice error from hapi. it doesn't return the error per-se, but the reply interface from hapi.

evq: we weren't calling reply previously in this block though

maikelmclauflin: right, we weren't calling it, so if there was a problem anywhere it was with the `wallet` not existing, and skipping the block in the try or via an error (say, from sentry) that was reported to sentry and then continuing on with the handler, returning the rates object only

evq: chatted out of band - need to check that expired token does not cause an exception

PR #585, opened 2019-03-20
abstracted env reading and most writing

### LINKED ISSUE #429: Centralize config env var readings
`process.env` reading and transforming should be centralized to `config.js` or another aptly named file to separate environment from usage and simplify inputs.

maikelmclauflin:
mostly fixes #429 

last issue to solve is how our project writes to process.env. haven't looked into why yet, but it may be necessary because of requirement of given package.


tsmartt:
Closing as stale.

PR #597, opened 2019-03-29
Wallet cooldown

maikelmclauflin:
fixes https://github.com/brave-intl/internal/issues/105


ledger/controllers/grants.js
```diff
@@ -1161,3 +1169,14 @@ function uploadTypedGrants (protocolVersion, uploadSchema, contentSchema) {
     reply({})
   }
 }
+
+function cooldownOffset () {
+  console.log('cooldown calc', typeof WALLET_COOLDOWN_HRS, WALLET_COOLDOWN_HRS)
```
nvonpentz: nit: log statement

ledger/controllers/grants.js
```diff
@@ -1161,3 +1169,14 @@ function uploadTypedGrants (protocolVersion, uploadSchema, contentSchema) {
...
+  const hours = WALLET_COOLDOWN_HRS ? (+WALLET_COOLDOWN_HRS || 0) : 24
+  return hours * 60 * 60 * 1000
+}
+
+function walletCooldownId () {
+  const hexString = braveUtils.hexTime(-cooldownOffset())
+  return bson.ObjectId(hexString)
+}
```
nvonpentz: could we unit test walletCooldownId and cooldownOffset?

bat-utils/lib/extras-utils.js
```diff
@@ -48,6 +49,12 @@ function documentOlderThan (olderThanDays, anchorTime, _id) {
   return createdTimestamp(_id) < (anchorTime - (DAY_MS * olderThanDays))
 }
 
+function hexTime (offset = 0, now = new Date()) {
+  const offsetMS = now.getTime() - offset
+  const offsetSeconds = Math.floor(offsetMS / 1000)
+  return offsetSeconds.toString(16) + '0000000000000000'
+}
+
```
nvonpentz: could we unit test hexTime?

maikelmclauflin:
> Adding an integration test would be great too if time permits

so i thought about this, i just couldn't think of any way to do it unless we want to create some sort of bypass token that we can use in tests. happy to do that if wanted

bat-utils/lib/extras-utils.js
```diff
@@ -48,6 +49,11 @@ function documentOlderThan (olderThanDays, anchorTime, _id) {
   return createdTimestamp(_id) < (anchorTime - (DAY_MS * olderThanDays))
 }
 
+function hexTime (now = new Date()) {
```
evq: I would prefer we call this something like `timestampToMongoId` or similar

maikelmclauflin: can do

maikelmclauflin: just realized it's not even used anymore, so removing

PR #595, opened 2019-03-27
Server isolation

maikelmclauflin:
fixes #594 

by moving the unit tests' servers to the same side running the tests, we were able to get a much better picture of the code that is covered

previous
```
File                           |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-------------------------------|----------|----------|----------|----------|-------------------|
All files                      |    23.95 |    15.96 |    22.72 |    25.17 |                   |
```

current
```
File                           |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-------------------------------|----------|----------|----------|----------|-------------------|
All files                      |    57.11 |    41.32 |    60.08 |    59.22 |                   |
```

PR #610, opened 2019-04-15
Create ads payout report and potential payments tables

nvonpentz:
Resolves https://github.com/brave-intl/gordon/issues/40

I'm not in love with the name `potential_payments_ads` but it does convey that these payments may not go through and it does mirror what's on publishers.

### DIFF
--- a/eyeshade/migrations/0006_payout_reports_ads/up.sql
+++ b/eyeshade/migrations/0006_payout_reports_ads/up.sql
@@ -1,0 +1,9 @@
+select execute($$
+
+insert into migrations (id, description) values ('0006', 'payout_reports_ads');
+
+create table payout_reports_ads(
+  id uuid primary key default uuid_generate_v4(),
+  created_at timestamp with time zone not null default current_timestamp
+)
+$$) where not exists (select * from migrations where id = '0006');
--- a/eyeshade/migrations/0006_payout_reports_ads/up.sql
+++ b/eyeshade/migrations/0006_payout_reports_ads/up.sql
@@ -1,5 +1,7 @@
 select execute($$
 
+create extension if not exists "uuid-ossp";
+
 insert into migrations (id, description) values ('0006', 'payout_reports_ads');
 
 create table payout_reports_ads(
--- a/eyeshade/migrations/0006_payout_reports_ads/up.sql
+++ b/eyeshade/migrations/0006_payout_reports_ads/up.sql
@@ -5,7 +5,7 @@ create extension if not exists "uuid-ossp";
 insert into migrations (id, description) values ('0006', 'payout_reports_ads');
 
 create table payout_reports_ads(
-  id uuid primary key default uuid_generate_v4(),
+  id         uuid primary key default uuid_generate_v4(),
   created_at timestamp with time zone not null default current_timestamp
 )
 $$) where not exists (select * from migrations where id = '0006');
--- a/eyeshade/workers/ads.js
+++ b/eyeshade/workers/ads.js
@@ -12,19 +12,15 @@ const createPotentialPaymentsQuery = `insert into potential_payments_ads (payout
 // and inserts them into potential_payments
 const monthly = async (debug, runtime) => {
   const client = await runtime.postgres.connect()
-  // const walletsCollection = runtime.database.get('wallets', debug)
-  const walletsCollection = await runtime.database.collection('wallets')
+  const walletsCollection = runtime.database.get('wallets', debug)
   const payoutReportId = uuidV4()
 
   try {
     await client.query('BEGIN')
-
     // First create the payout report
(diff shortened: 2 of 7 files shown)


eyeshade/migrations/0006_payout_reports_ads/up.sql
```diff
@@ -0,0 +1,11 @@
+select execute($$
+
+create extension if not exists "uuid-ossp";
```
nvonpentz: This allows postgres to generate uuids for us instead of doing it in the app.

eyeshade/workers/ads.js
```diff
@@ -0,0 +1,48 @@
...
+const selectWalletBalancesQuery = `select account_id, balance from account_balances where account_type = 'payment_id'`
+
+const createPotentialPaymentsQuery = `insert into potential_payments_ads (payout_report_id, payment_id, provider_id, amount) values ($1, $2, $3, $4)`
+
+// Takes a snapshot of ad account balances
+// and inserts them into potential_payments
+const monthly = async (debug, runtime) => {
+  const client = await runtime.postgres.connect()
+  // const walletsCollection = runtime.database.get('wallets', debug)
+  const walletsCollection = await runtime.database.collection('wallets')
```
nvonpentz:
```js
// const walletsCollection = runtime.database.get('wallets', debug)
const walletsCollection = await runtime.database.collection('wallets')
```
I think the commented out line is the one I want to use and that the current one will fail in prod.  However, I don't know how to pass a runtime object with a database object that would result in the wallets collection being defined in the test [here](https://github.com/brave-intl/bat-ledger/pull/610/files#diff-70cf2a668842f4ac48e3542aaff8cd2cR48).

maikelmclauflin: create a runtime-database object or use the boot-runtime module like i do here: https://github.com/brave-intl/bat-ledger/pull/595/files#diff-ff2899c03c9e8e31a9b69d7621e08c83R264

maikelmclauflin: https://github.com/brave-intl/bat-ledger/blob/master/test/ledger/surveyor.integration.test.js#L23

eyeshade/migrations/0006_payout_reports_ads/up.sql
maikelmclauflin: i like it

test/eyeshade/ads.integration.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+  // Run the payout report
+  await monthly({}, { postgres: postgres, database: eyeshadeMongo })
+
+  // Ensure the wallet balance made it in
+  const potentialPayments = (await postgres.query(`select * from potential_payments_ads`)).rows
+  t.true(potentialPayments.length === 1)
+
+  const potentialPayment = potentialPayments[0]
+  t.true(potentialPayment.payment_id === paymentId)
```
maikelmclauflin: I would actually like to start moving toward using ava's `is` method wherever possible, and adding in another param for the message like so 
```
t.is(potentialPayment.payment_id, paymentId, 'the inserted payment id matches')
```

eyeshade/workers/ads.js
maikelmclauflin: force

test/eyeshade/ads.integration.test.js
```diff
@@ -0,0 +1,68 @@
...
+})
+
+test.afterEach.always(async t => {
+  await cleanPgDb(postgres)()
+  await cleanDbs()
+})
+
+test('ads payout report cron job takes a snapshot of balances', async t => {
+  const eyeshadeMongo = await connectToDb('ledger')
+  const wallets = eyeshadeMongo.collection('wallets')
```
maikelmclauflin: can we just use the runtime db here?

nvonpentz: Probably, I will check and revert.  Also need to update the variable name since it's no longer eyeshadeMongo

test/eyeshade/ads.integration.test.js
```diff
@@ -0,0 +1,68 @@
...
+  const txId = uuidV4()
+  const createdAt = new Date()
+  const description = 'funding tx for viewing ads'
+  const transactionType = 'ad'
+  const fromAccountType = 'uphold'
+  const fromAccount = uuidV4()
+  const toAccountType = 'payment_id'
+  const amount = 1000
+
+  const insertQuery = `insert into transactions (id, created_at, description, transaction_type, from_account_type, from_account, to_account_type, to_account, amount) values ($1, $2, $3, $4, $5, $6, $7, $8, $9)`
```
maikelmclauflin: if you'd like, there's an abstraction for this in eyeshade/lib/transaction

PR #611, opened 2019-04-15
merged prometheus registries across dynos

### LINKED ISSUE #609: implement prometheus instrumentation for uphold requests
rewrite runtime-wallet to create an interface for uphold's api method to capture all requests

https://uphold.github.io/uphold-sdk-javascript/client.html

maikelmclauflin:
fixes #609


bat-utils/lib/runtime-uphold.js
```diff
@@ -0,0 +1,65 @@
...
+      cardinality
+    }
+  }
+  async request (url, method, body, customHeaders = {}) {
+    const { prometheus } = this
+    const {
+      path,
+      cardinality
+    } = this.path(url)
+    const name = 'upholdApi_request_buckets_milliseconds'
```
maikelmclauflin: @evq is there a better pattern than this? / is snake case all around preferable? i just went with camel for the first part of the name so to allow for multiword, but kebab would work too, no?

package.json
```diff
@@ -92,6 +92,7 @@
     "hapi-swagger": "github:brave/hapi-swagger",
     "http-request-signature": "0.0.2",
     "inert": "^4.2.1",
+    "isuuid": "^0.1.0",
```
diracdeltas: I rarely say this, but I think it would be better if you just copy-pasted the 2 lines from the repository instead of adding it as a dependency: https://github.com/nescalante/isuuid/blob/master/index.js#L6-L8

Rationale:
* It's only 2 lines, can be combined into 1 line easily.
* The package hasn't been updated since 2015
* Adding a package dependency adds some risk that it'll be compromised/broken, like what happened with leftpad. https://www.theregister.co.uk/2016/03/23/npm_left_pad_chaos/

maikelmclauflin: happy to!

bat-utils/lib/runtime-uphold.js
evq: yeah I think all snake is more canonical. maybe just `uphold_request_buckets_milliseconds`?

maikelmclauflin: ah, right. yeah, i guess i don't need the other ones now because its all going through the uphold api

ledger/controllers/registrar.js
```diff
@@ -161,34 +161,36 @@ const createPersona = function (runtime) {
...
-          const validity = verify({
-            headers: request.payload.request.headers,
-            publicKey: request.payload.request.body.publicKey
-          }, {
-            algorithm: 'ed25519'
-          })
-          if (!validity.verified) {
-            throw boom.badData('wallet creation request failed validation, http signature was not valid')
-          }
+        const validity = verify({
```
evq: this isn't anonize verification

PR #614, opened 2019-04-24
added type to grant content subset

### LINKED ISSUE #613: type missing from grants object
make sure type key and value is in grants object, or at least not removed

`balance://v2/wallet/{paymentId}/balance`

maikelmclauflin:
fixes #613


ledger/controllers/wallet.js
```diff
@@ -163,7 +163,9 @@ async function sumActiveGrants (runtime, info, wallet, grants) {
     } else {
       let content = braveUtils.extractJws(token)
       total = total.plus(content.probi)
-      results.push(underscore.pick(content, ['altcurrency', 'expiryTime', 'probi']))
+      const exposedContent = underscore.pick(content, ['altcurrency', 'expiryTime', 'probi', 'type'])
+      exposedContent.type = exposedContent.type || 'ugp'
```
maikelmclauflin: had an awful time with this bit. for some reason i couldn't get the type to flow through even with the joi validation applied.

evq: ah I assume you tried `optional()`?

maikelmclauflin: i did not, but when i tried the setup on runkit i was getting the correct result

PR #631, opened 2019-05-30
Fastly Integration

maikelmclauflin:
may not be necessary but leaving here for now until we can fix the prometheus going directly to heroku app


bat-utils/lib/hapi-auth-whitelist.js
```diff
@@ -35,11 +37,17 @@ exports.ipaddr = (request) => {
   // Since it is easy to forge an X-Forwarded-For field the given information should be used with care.
   // The last IP address is always the IP address that connects to the last proxy, which means it is the most reliable source of information.
 
-  const forwardedFor = request.headers['x-forwarded-for']
+  const { headers } = request
+  const forwardedFor = headers['x-forwarded-for']
+  const token = headers['fastly-token']
   if (forwardedFor) {
-    const shift = forwardedIPShift()
+    if (fastlyTokens.length && !braveHapi.isSimpleTokenValid(fastlyTokens, token)) {
```
evq: can we fail if we're in production and the env var is not passed?

evq:
can you update the issue name?

bat-utils/lib/hapi-auth-whitelist.js
maikelmclauflin: no, defaults to an empty array above: https://github.com/brave-intl/bat-ledger/pull/631/files#diff-d3f10f89496dc0476f73b80d818352e5R9

evq: sorry I wasn't clear. can you please make it fail in that case

maikelmclauflin: yeah, sure

PR #642, opened 2019-06-12
pointed to ads service to load list of country codes where ads are available

maikelmclauflin:
in order to keep the list of ads available up to date, adjusted to load the list of country codes.


ledger/controllers/grants.js
```diff
@@ -232,7 +232,7 @@ const getGrant = (protocolVersion) => (runtime) => {
     entries = await promotions.find(query)
     if ((!entries) || (!entries[0])) return reply(boom.notFound('no promotions available'))
 
-    const adsAvailable = adsGrantsAvailable(request.headers['fastly-geoip-country-code'])
+    const adsAvailable = await adsGrantsAvailable(request.headers['fastly-geoip-country-code'])
```
evq: I don't think there's any reason to look up the ads region list on every single request. What about loading once on startup?

maikelmclauflin: oh, yeah, no it only loads once on startup, then it just keeps resolving the same promise.

PR #648, opened 2019-06-24
closed test's loop and added transaction type

### LINKED ISSUE #647: accept manual_settlement transaction types in accounts endpoint
`GET /v1/accounts/{account}/transactions`
does not currently validate correctly when `manual_settlement`s are present for the account. need to update validator to allow those types to pass through

maikelmclauflin:
fixes #647

### DIFF
--- a/eyeshade/controllers/accounts.js
+++ b/eyeshade/controllers/accounts.js
@@ -16,7 +16,7 @@ const v1 = {}
 
 const settlementDestinationTypes = ['uphold']
 const accountTypes = ['channel', 'owner'].concat(settlementDestinationTypes)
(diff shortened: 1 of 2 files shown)


eyeshade/controllers/accounts.js
```diff
@@ -16,7 +16,7 @@ const v1 = {}
 
 const settlementDestinationTypes = ['uphold']
 const accountTypes = ['channel', 'owner'].concat(settlementDestinationTypes)
-const transactionTypes = ['contribution', 'referral', 'contribution_settlement', 'referral_settlement', 'fees', 'scaleup', 'manual', 'user_deposit', 'manual_payout']
+const transactionTypes = ['contribution', 'referral', 'contribution_settlement', 'referral_settlement', 'fees', 'scaleup', 'manual', 'user_deposit', 'manual_payout', 'manual_settlement']
```
evq: I think we can remove `manual_payout` that type doesn't exist

PR #651, opened 2019-06-27
refresh view outside of transaction

maikelmclauflin:


eyeshade/workers/publishers.js
```diff
@@ -36,20 +36,22 @@ exports.workers = {
...
-          runtime.captureException(e, { extra: { report: 'settlement-report', settlementId } })
-          throw e
+        for (let doc of docs) {
+          await insertFromSettlement(runtime, client, doc)
         }
-
+        await client.query('COMMIT')
         if (shouldUpdateBalances) {
-          await updateBalances(runtime, client, true)
+          updateBalances(runtime, client, true)
```
evq: shouldn't this be moved after our `catch`? because the transaction is already committed so there's no point to trying to roll back

PR #654, opened 2019-06-27
Fee checker

maikelmclauflin:


evq:
per our chat, due to concerns raised in security review we will fetch p2p tip information directly from uphold in the pub app

PR #658, opened 2019-07-02
removed all refresh balances

### LINKED ISSUE #656: votes not propagating
voting is not propagating through workers and databases.
short term fix is to reduce the number of times db must refresh (causing memory issues)
long term fix is to get pr #524 in so that jobs are not being dropped

maikelmclauflin:
fixes #656

### DIFF
--- a/bat-utils/lib/youtube.test.js
+++ b/bat-utils/lib/youtube.test.js
@@ -1,6 +1,8 @@
 'use strict'
 import test from 'ava'
 import getYoutubeChannelId from './youtube'
+import dotenv from 'dotenv'
+dotenv.config()
 
 test.skip('retrieves the youtube channel id for a youtube username', async t => {
   const youtubeUsername = 'SaturdayNightLive'
--- a/bat-utils/lib/youtube.test.js
+++ b/bat-utils/lib/youtube.test.js
@@ -1,8 +1,6 @@
 'use strict'
 import test from 'ava'
 import getYoutubeChannelId from './youtube'
-import dotenv from 'dotenv'
-dotenv.config()
 
 test.skip('retrieves the youtube channel id for a youtube username', async t => {
   const youtubeUsername = 'SaturdayNightLive'
--- a/eyeshade/workers/publishers.js
+++ b/eyeshade/workers/publishers.js
@@ -1,4 +1,4 @@
-const { insertFromSettlement, updateBalances } = require('../lib/transaction.js')
+const { insertFromSettlement } = require('../lib/transaction.js')
 
 exports.name = 'publishers'
 exports.initialize = async (debug, runtime) => {
@@ -10,8 +10,7 @@ exports.workers = {
 
     { queue   : 'settlement-report'
     , message :
-      { shouldUpdateBalances : false,
-        settlementId         : '',
+      { settlementId         : '',
         type                 : '',
       }
     }
@@ -20,7 +19,6 @@ exports.workers = {
     async (debug, runtime, payload) => {
       const settlementsCollection = runtime.database.get('settlements', debug)
       const {
-        shouldUpdateBalances,
         settlementId,
         type
       } = payload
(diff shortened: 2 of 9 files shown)


bat-utils/lib/youtube.test.js
```diff
@@ -1,6 +1,8 @@
 'use strict'
 import test from 'ava'
 import getYoutubeChannelId from './youtube'
+import dotenv from 'dotenv'
```
evq: why do we need this?

eyeshade/workers/publishers.js
```diff
@@ -46,9 +44,6 @@ exports.workers = {
           throw e
         }
         await client.query('COMMIT')
```
evq: settlement reports run very infrequently (~2 times a month). I think we should keep the refresh here

eyeshade/workers/reports.js
```diff
@@ -74,5 +77,16 @@ exports.initialize = async (debug, runtime) => {
 
   if ((typeof process.env.DYNO === 'undefined') || (process.env.DYNO === 'worker.1')) {
     setTimeout(() => { daily(debug, runtime) }, 5 * 1000)
+    updateBalancesContinuously(runtime)
   }
 }
+
+async function updateBalancesContinuously (runtime) {
```
evq: I think continuous refresh is too aggressive

bat-utils/lib/youtube.test.js
maikelmclauflin: i found that i have to source the env to prevent the test from failing unless this line is in. if i recall correctly, ava starts a new process, so if it's not ones' env already (with `npm t`) test fails

eyeshade/workers/publishers.js
maikelmclauflin: i mean, yes, but it happens continuously

eyeshade/workers/reports.js
maikelmclauflin: perhaps a mechanism to keep it limited at 1 running at a time, that does not run if it is not asked to?

evq: I think I'm partial to something more like updating in the following two cases:
1. A settlement report was uploaded
2. Every 6 hours

maikelmclauflin: ok. yeah, that can be done

eyeshade/workers/reports.js
```diff
@@ -74,5 +77,14 @@ exports.initialize = async (debug, runtime) => {
 
   if ((typeof process.env.DYNO === 'undefined') || (process.env.DYNO === 'worker.1')) {
     setTimeout(() => { daily(debug, runtime) }, 5 * 1000)
+    updateBalancesOnInterval(runtime)
   }
 }
+
+async function updateBalancesOnInterval (runtime) {
+  await updateBalances(runtime)
+  const now = +(new Date())
```
nvonpentz: Any reason to do this over `(new Date()).getTime()`?

maikelmclauflin: habit i suppose. i used to write really tersely. they all do the same thing
```
(new Date()).valueOf()
```

eyeshade/workers/reports.js
```diff
@@ -74,5 +77,14 @@ exports.initialize = async (debug, runtime) => {
...
+  const hours6 = 1000 * 60 * 60 * 6
+  const msUntilNext = now % hours6
+  setTimeout(() => updateBalancesOnInterval(runtime), msUntilNext)
+}
```
nvonpentz: I'm not sure this does what you want here.  I think what you're after is scheduling things when the hour % 6 == 0, so the 0th hour of the day, 6th, 12th and so on.

But I don't think we want to set the timeout to the remainder.  Suppose it is 6:01AM.  The remainder would be 1 minute, and it'd get scheduled to run in 1 minute.  That might finish and it's 6:15, at which point we schedule the next job to run in 15 minutes.

Does that make sense or perhaps I'm missing something?

maikelmclauflin: ah, i am missing a line aren't i. yeah, i'll update

PR #671, opened 2019-07-22
added instrumentation to all workers

maikelmclauflin:
instruments duration for workers so that we can get more information about bottlenecks / downed / erring workers


tsmartt:
Closing as stale.

PR #674, opened 2019-07-23
claimed with 0 as amount

maikelmclauflin:
allows user to claim their wallet with a 0 amount in transaction

### DIFF
--- a/test/e2e.integration.test.js
+++ b/test/e2e.integration.test.js
@@ -512,20 +512,22 @@ test('wallets can be claimed by verified members', async (t) => {
   const settlement = process.env.BAT_SETTLEMENT_ADDRESS
 
   await claimCard(anonCardInfo1, settlement)
+  await claimCard(anonCardInfo2, anonCardInfo1.providerId, 200, '0')
   await claimCard(anonCardInfo2, anonCardInfo1.providerId)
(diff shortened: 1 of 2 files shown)


test/e2e.integration.test.js
```diff
@@ -512,20 +512,22 @@ test('wallets can be claimed by verified members', async (t) => {
...
+  await claimCard(anonCardInfo2, anonCardInfo1.providerId, 200, '0')
   await claimCard(anonCardInfo2, anonCardInfo1.providerId)
   await claimCard(anonCardInfo3, anonCardInfo2.providerId)
   await claimCard(anonCardInfo4, anonCardInfo3.providerId, 409)
 
-  // redundant calls are fine
-  await claimCard(anonCardInfo1, settlement)
+  // redundant calls are fine if they do not try to transfer an amount
+  await claimCard(anonCardInfo3, settlement, 200, '0')
+  await claimCard(anonCardInfo3, settlement, 500)
```
evq: what?

evq: this section doesn't make sense. redundant calls are fine provided the amount we are attempting to transfer is less than the balance.

also we probably shouldn't return 500 in this case

evq: clarified / removed test. created #675 to fix the 500

PR #682, opened 2019-07-25
Fix ads payout report generation

### LINKED ISSUE #678: automatically generate ads payout report every month
I attempted to do this in https://github.com/brave-intl/bat-ledger/pull/610 however was unsuccessful.  The payout report is not generated on a consistent timeline on staging, and not generated at all on production.

We should figure out how to run this code on a consistent interval by either fixing #610 or by coming up with a new method to schedule jobs (heroku scheduler, new job framework, etc).

nvonpentz:
Resolves #678 

Builds on https://github.com/brave-intl/bat-ledger/pull/610

The first commit fixes two things about the query that selects the balances to be converted into ads potential payments

1. Adds an extra constraint that the balance is greater than 0. This is a constraint on the potential_payments table, this needs to be true anyway or else the entire transaction will be aborted (which we've seen on prod).

2. This ensures that we are only paying out transactions up to the first moment of the first of the month. All balances will be taken at the same time.

The second commit adds extra constraints to the potential_payments_ads table.  I added this to target a problem that I later realized we aren't having, but it would be good to add anyway.

### DIFF
--- a/eyeshade/workers/ads.js
+++ b/eyeshade/workers/ads.js
@@ -4,7 +4,21 @@ const underscore = require('underscore')
 
 const createPayoutReportQuery = `insert into payout_reports_ads (id) values ($1)`
 
-const selectWalletBalancesQuery = `select account_id, balance from account_balances where account_type = 'payment_id'`
+const selectWalletBalancesQuery = `
+  with ads_balances as (
+    select
+      account_id,
+      sum(amount) as balance,
+    from account_transactions
+    where account_type = 'payment_id'
+    and created_at < date_trunct('month', current_date)
+  )
+  select
+    account_id,
+    balance
+  from ads_balances
+  where balance > 0
+`
 
 const createPotentialPaymentsQuery = `insert into potential_payments_ads (payout_report_id, payment_id, provider_id, amount) values ($1, $2, $3, $4)`
 
--- a/eyeshade/workers/ads.js
+++ b/eyeshade/workers/ads.js
@@ -12,6 +12,8 @@ const selectWalletBalancesQuery = `
     from account_transactions
     where account_type = 'payment_id'
     and created_at < date_trunc('month', current_date)
+    group by account_id
+    order by balance desc
   )
   select
     account_id,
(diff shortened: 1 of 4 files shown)


nvonpentz:
Weird not all the commits on this branch are being updated on the pull request..

nvonpentz:
We can add more constraints so we don't duplicate reports later if we desire, but this should be good for now

eyeshade/workers/ads.js
```diff
@@ -4,13 +4,31 @@ const underscore = require('underscore')
...
+  where balance > 0
+`
 
 const createPotentialPaymentsQuery = `insert into potential_payments_ads (payout_report_id, payment_id, provider_id, amount) values ($1, $2, $3, $4)`
 
 // Takes a snapshot of ad account balances
 // and inserts them into potential_payments
 const monthly = async (debug, runtime) => {
+  // Limit the dynos that can run this worker to 1
+  if ((typeof process.env.DYNO !== 'undefined') && (process.env.DYNO !== 'worker.1')) return
```
maikelmclauflin: would you mind moving this into the initialize fn? would be nice to not have to think about these requirements when calling monthly if other things around it change.

test/eyeshade/ads.integration.test.js
```diff
@@ -34,13 +34,11 @@ test('ads payout report cron job takes a snapshot of balances', async t => {
...
   const providerId = uuidV4()
   await wallets.insert({ paymentId, providerId })
-  const insertedWallet = await wallets.findOne({ paymentId: paymentId })
-  console.log('inserted wallet is')
-  console.log(insertedWallet)
 
   // Create an ad transaction so there is a payment_id with a balance
   const txId = uuidV4()
   const createdAt = new Date()
+  createdAt.setMonth(createdAt.getMonth() - 2)
```
maikelmclauflin: this does not set the date to `2019-02-01T:00:00:00.000Z` it only modifies the month, so we would have a gap at between each payout depending on if one failed or if one was late or otherwise

PR #683, opened 2019-07-25
stored prometheus on redis

maikelmclauflin:
store prometheus on redis


evq:
@maikelmclauflin looks like this needs a rebase

PR #685, opened 2019-07-26
Automatically add balancing transactions when uploading ads grants

### LINKED ISSUE #679: provide easy way to enter balancing ads transactions after ads settlement
After we sign the ads grants and upload to the ledger, we need to enter the transactions to show that we've sent the BAT to the wallets to avoid double spending.

nvonpentz:
Resolves #679 

The proposed behavior is such: when the ads grants are signed and uploaded, pass a message to the shared redis queue.  The message is picked up by an eyeshade worker and the transaction is automatically inserted.

One problem is that grants are not active by default.  We should really only be inserting transactions for active grants.

### DIFF
--- a/eyeshade/workers/ads.js
+++ b/eyeshade/workers/ads.js
@@ -43,3 +43,17 @@ exports.initialize = async (debug, runtime) => {
 }
 
 exports.monthly = monthly
+
+exports.workers = {
+/* sent by ledger POST /v4/grants
+ */
+  'settlement-report-ads':
+    async (debug, runtime, payload) => {
+      const { postgres } = runtime
+
+      // TODO
+      // for each of the providerId, amount pairs
+      // that we made grants available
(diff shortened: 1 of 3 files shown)


nvonpentz:
@evq @maikelmclauflin Let me know what you think of this approach.

maikelmclauflin:
another possible alternative would be to do it at claim time

nvonpentz:
I like doing it at claim time but I think we would run into a double spend scenario if someone doesn't claim their grant and gets another next month.  We'd have to restrict to only one active grant at a t time.

@evq suggests not tying it to existing grant stuff in bat-ledger since it will be replaced in the grant rework.

I'm okay with doing this manually until grant rework is finished.

evq:
@nvonpentz just doing a bit of cleanup, it seemed like there wasn't any implementation here - let me know if we should reopen

PR #701, opened 2019-08-14
added counter for referrals

### LINKED ISSUE #575: use prometheus to monitor that referral confirmations are continually happening

maikelmclauflin:
fixes #575

### DIFF
--- a/eyeshade/controllers/referrals.js
+++ b/eyeshade/controllers/referrals.js
@@ -95,6 +95,7 @@ v1.createReferrals = {
         entries.forEach((referral) => { existingDownloadIds.push(referral.downloadId) })
       }
 
+      let insertedReferrals = 0
       for (let referral of payload) {
         let state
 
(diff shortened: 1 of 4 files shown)


eyeshade/controllers/referrals.js
```diff
@@ -115,9 +116,11 @@ v1.createReferrals = {
           }, underscore.pick(referral, [ 'platform', 'altcurrency', 'probi' ]))
         }
         await referrals.update({ downloadId: referral.downloadId }, state, { upsert: true })
+        insertedReferrals += 1
       }
-
       await runtime.queue.send(debug, 'referral-report', { transactionId })
+      console.log('received referral')
```
evq: I think we already log the queue send - do we need this?

eyeshade/workers/referrals.js
```diff
@@ -53,6 +53,8 @@ exports.workers = {
           throw e
         }
         await client.query('COMMIT')
+        console.log('inserted referral')
```
evq: same as above, we already log the queue receive - right?

PR #709, opened 2019-08-20
delete owners

evq:
depends on a few remaining publishers side changes


maikelmclauflin:
confirmed that `GET /v1/owners/{owner}/wallet` is good to remove

Issue #712, opened 2019-08-27
Use install geo when calculating value of a finalized referral install


evq:
Currently we pay a flat $5 in BAT but we would like to move toward having different "rate cards" for various countries. We'll need to modify the endpoint that the referral promo server uses to report finalized installs as well as the logic that actually calculates the BAT value of finalized installs.

PR #713, opened 2019-08-27
Look up referral amount by location

azach:
This adds geo-specific amounts to referrals based on their country code location. A new table, `geo_referral_amounts` is created to store these values. If an amount isn't found for a given location, we fall back to the currently specified default.


maikelmclauflin:
fixes #712

maikelmclauflin:
to be merged in at #724

PR #715, opened 2019-08-28
reversed computation for metric

maikelmclauflin:


bat-utils/lib/runtime-prometheus.js
```diff
@@ -145,7 +145,7 @@ function registerMetrics (prometheus) {
   register.registerMetric(viewRefreshRequestBucketsMilliseconds)
 
   const settlementCounter = new client.Counter({
-    name: 'settlement_balance_counter',
+    name: 'funds_received_count',
     help: 'a count up of the number of bat removed from the settlement wallet'
```
evq: we should probably update this too

PR #716, opened 2019-08-29
Snapshot endpoint

### LINKED ISSUE #718: stats endpoint for pubs
get a daily snapshot of stats for publishers. triages: https://github.com/brave-intl/publishers/issues/2132

maikelmclauflin:
created endpoint for retrieving created snapshots
fixes #718


eyeshade/migrations/0010_snapshots/up.sql
```diff
@@ -0,0 +1,12 @@
+select execute($$
+
+insert into migrations (id, description) values ('0010', 'snapshots');
+
+create table snapshots(
+  id               uuid primary key default uuid_generate_v4(),
```
yachtcaptain23: So it seems you're merging the snapshot data together for `votes` and `transactions`. I'd suggest you split out data by its type by creating another column `kind` or `type` so when we query it'll just query the right rows rather than including excess data.

maikelmclauflin: added that feature. but pretty sure we can do that by just reaching in

evq:
waiting on some feedback from @yachtcaptain23 on the approach

evq:
closing as we'll be caching long running queries elsewhere

PR #717, opened 2019-09-04
bonus payouts

maikelmclauflin:
added hash and logic for bonus accounting


test/ledger/wallet.integration.test.js
```diff
@@ -170,7 +176,7 @@ test('compositing wallet grant information', async (t) => {
   } = await ledgerAgent.get(`/v2/wallet/${paymentId}/grants/ads`).expect(ok)
   const expectedAds = createComposite({
     type: 'ads',
-    amount: (new BigNumber(2)),
+    amount: (new BigNumber(2)).plus(13),
```
nvonpentz: why do we add 13 here?

maikelmclauflin: that's for the bonus. the bonus is `25` and the value in the grant is `38` so 38-25=13

maikelmclauflin: so it's as if they earned 25 in the bonus and 13 from doing the work of clicking on ads

PR #720, opened 2019-09-13
allow optional publisher parameter in settlement report

evq:


eyeshade/workers/publishers.js
```diff
@@ -25,16 +27,23 @@ exports.workers = {
...
+      const query = {
         type,
         settlementId,
         owner: {
           $exists: true
         }
-      })
+      }
+
+      if (publisher) {
```
maikelmclauflin: i don't see publisher passed from anywhere. is this supposed to be triggered from an external service?

evq: yeah this is e.g. if we want to run a manual script that enters in specific settlement transactions that were missed in the original aggregate job

PR #724, opened 2019-09-19
Geo referral amounts

maikelmclauflin:


eyeshade/migrations/0010_geo_referral/up.sql
```diff
@@ -0,0 +1,25 @@
+select execute($$
+
+insert into migrations (id, description) values ('0010', 'geo_referral');
+
+create table geo_referral_groups(
+  id                    uuid primary key default uuid_generate_v4(),
+  created_at            timestamp with time zone not null default current_timestamp,
+  min_referral_time     timestamp with time zone not null default current_timestamp,
```
nvonpentz: What is `min_referral_time`?  Is it possible to rename it to something ending in `_at`?

eyeshade/lib/queries.js
```diff
@@ -65,3 +67,35 @@ function timeConstraintSettlements (options = {}) {
...
+  SELECT
+    group_id,
+    array_agg(country_code) AS codes
+  FROM geo_referral_countries
+  GROUP BY group_id
+) AS countries
+WHERE
+    countries.group_id = geo_referral_groups.id${_.isBoolean(active) ? `
+AND active = ${active}` : ''};`
+}
```
nvonpentz: Nice `array_agg`

eyeshade/lib/queries.js
```diff
@@ -65,3 +67,35 @@ function timeConstraintSettlements (options = {}) {
...
+SELECT
+  id,
+  min_referral_time as "minReferralTime"
+  ${fields.reduce((memo, key) => {
+    const query = fieldsMap[key]
+    if (query) {
+      memo.push(',\n  ', query)
+    }
+    return memo
+  }, []).join('')}
```
nvonpentz: Would it be simpler to just request all the fields and return null if they don't exist than dynamically compute the query?

test/utils.js
```diff
@@ -259,3 +274,42 @@ function makeSettlement (type, balance, overwrites = {}) {
     hash: uuidV4()
   }, overwrites)
 }
+
+async function insertReferralInfos (client) {
+  const ratesPaths = [{
+    path: filePath('0010_geo_referral', 'groups.csv'),
```
nvonpentz: You can also keep the seed data in an insert query so you don't have to keep a csv file, etc if you prefer

eyeshade/migrations/0010_geo_referral/up.sql
maikelmclauflin: so, min referral time is the cutoff time for download timestamps that the promo server uses.
we could change this into active_at or something like that.

eyeshade/lib/queries.js
maikelmclauflin: 😄 thanks. if there is a better way to do it though this i'm all for it

eyeshade/lib/queries.js
maikelmclauflin: i suppose there is no concern about the size of the response so yeah, i'd be fine with that

eyeshade/controllers/referrals.js
```diff
@@ -1,23 +1,68 @@
...
+const extrasUtils = utils.extras.utils
+
+const queries = require('../lib/queries')
 
 const v1 = {}
 
-const schema = Joi.array().min(1).items(Joi.object().keys({
+const amountValidator = braveJoi.string().numeric()
+const groupNameValidator = Joi.string().optional().description('the name given to the group')
+const channelIdValidator = braveJoi.string().publisher().required().description('the publisher identity')
```
corymcdonald: I think `publisherIdValidator` might be a more opt name for this since we're validating the publisher identity and not a channel

eyeshade/controllers/referrals.js
```diff
@@ -1,23 +1,68 @@
...
+
+const anyReferralVersion = Joi.alternatives().try(
+  manyGroupStampedReferrals,
+  manyReferrals
+)
+
+const referralGroupCountriesValidator = Joi.object().keys({
+  id: Joi.string().guid().required().description('the group id to report back for correct value categorization'),
+  activeAt: Joi.date().iso().required().description('the download cut off time to honor the amount'),
+  name: groupNameValidator.optional().description('name of the group'),
```
corymcdonald: I would expect a name would be required for a group as that's what we're likely going to display to the end user

maikelmclauflin: so, this is a fair point, but ledger is serving multiple services with this endpoint. it may make more sense to make activeAt optional, since everyone needs the id and the cross section of needs of fields is / will be different for everyone

test/eyeshade/referrals.integration.test.js
```diff
@@ -111,7 +157,113 @@ test('if promo sends mix of duplicate and valid referrals with same tx id, only
...
+}
+
+async function sendReferral (timestamp, groupId) {
+  const txId = uuidV4().toLowerCase()
+  const referral = {
+    downloadId: uuidV4().toLowerCase(),
+    channelId: braveYoutubePublisher,
+    platform: 'ios',
+    finalized: timestamp || new Date(),
+    // should attribute 6.5 bat / referral
```
evq: is this true? shouldn't it depend on the `groupId`

test/utils.js
```diff
@@ -259,3 +273,26 @@ function makeSettlement (type, balance, overwrites = {}) {
...
+  for (const subject of ratesPaths) {
+    await insertCSVFileToDB(client, subject.path)
+  }
+
+  function filePath (...paths) {
+    return path.join(__dirname, '..', 'eyeshade', 'migrations', ...paths)
+  }
+}
+
+function insertCSVFileToDB (client, path) {
```
evq: this is taking a sql file, not a csv

PR #725, opened 2019-09-25
Hapi upgrade

### LINKED ISSUE #257: investigate upgrade to hapi v17

maikelmclauflin:
~need https://github.com/brave-intl/bat-go/pull/87 merged first~
fixes #257


maikelmclauflin:
not sure why this is failing yet, but it seems to have something to do with how things are being bound on network level, specifically this line: https://github.com/brave-intl/bat-ledger/pull/725/files#diff-52adb75d0ebd9a9f61c97346b2a7b7c1R50

bat-utils/lib/hapi-auth-whitelist.js
```diff
@@ -22,20 +22,45 @@ const internals = {
   implementation: (server, options) => { return { authenticate: exports.authenticate } }
 }
 
-exports.authorizedP = (ipaddr) => {
+exports.authorizedP = authorizedP
+// NOTE This function trusts the final IP address in X-Forwarded-For
```
evq: can you move this down to where the function is defined

bat-utils/lib/hapi-rate-limiter.js
```diff
@@ -0,0 +1,147 @@
...
+    if ((entry.indexOf('/') === -1) && (entry.split('.').length === 4)) return graylist.authorizedAddrs.push(entry)
+
+    graylist.authorizedBlocks.push(new Netmask(entry))
+  })
+}
+
+module.exports = (runtime) => {
+  const redisClient = (runtime.cache && runtime.cache.cache) || runtime.queue.config.client
+
+  // const redisClient = redis.createClient({
```
evq: can we remove this?

bat-utils/lib/hapi-controllers-login.js
```diff
@@ -17,42 +15,34 @@ const v1 = {}
...
-            ': ' + JSON.stringify(credentials.scope) + ' at ' + os.hostname() + ' ' + npminfo.name + '@' +
-            npminfo.version + (process.env.DYNO ? ' at ' + process.env.DYNO : '') + ' from ' + whitelist.ipaddr(request)
-        })
-
-        request.cookieAuth.set(credentials)
-        reply.redirect(runtime.login.github.world)
+    try {
+      await checkTeams(runtime, credentials)
+    } catch (e) {
+      return runtime.notify(debug, {
```
evq: should we be returning `runtime.notify` or re-throwing `e`?

bat-utils/lib/hapi-controllers-login.js
```diff
@@ -109,6 +100,27 @@ v1.logout = {
...
+
+  const { data: teams } = await octokit.teams.listForAuthenticatedUser({
+    org: organization
+  })
+
+  if (!teams.find(({ organization: { login } }) => login === organization)) {
+    throw boom.forbidden()
+  }
+  return {
+    scope: [organization]
```
evq: the scope here is different than it was before - previously the scope was one or more team names, not the organization :/

how can this be working?

maikelmclauflin: scope is not currently used. i just returned it in case we wanted to write a test or something else

bat-utils/lib/hapi-controllers-login.js
maikelmclauflin: rethrowing e. good catch

bat-utils/lib/hapi-server.js
```diff
@@ -66,194 +74,112 @@ const Server = async (options, runtime) => {
...
-      whitelist,
+      hapiCookie,
+      {
+        plugin: whitelist.plugin
+      },
       inert,
       vision,
+      rateLimiter(runtime),
+      swagger()
+    ], process.env.NODE_ENV === 'production' ? [
```
jseidl: I'd also enforce HTTPs for `staging`

bat-utils/lib/hapi-server.js
```diff
@@ -66,194 +74,112 @@ const Server = async (options, runtime) => {
...
-        debug('unable to register extensions', err)
-        reject(err)
-        throw err
-      }
+      })
 
+      debug('session authentication strategy via cookie')
+    } else {
+      debug('github authentication disabled')
       if (process.env.NODE_ENV === 'production') {
```
jseidl: IMO all checks for `production` should check not `development` since I believe `staging` to be `the next production` and we need to make sure it will behave as in `production`

bat-utils/lib/hapi-server.js
evq: somewhat confusingly we also set `NODE_ENV` to production on staging :)

bat-utils/lib/hapi-server.js
jseidl: Resolved by @evq comment above

balance/controllers/address.js
```diff
@@ -41,7 +41,7 @@ module.exports.configuration = {
 
 v2.walletBalance =
 { handler: (runtime) => {
-  return async (request, reply) => {
+  return async (request, h) => {
```
diracdeltas: in all these functions where `h` is unused, is there any particular reason to include it at all?

evq: consistency with the expected handler function signature

bat-utils/lib/hapi-rate-limiter.js
```diff
@@ -0,0 +1,149 @@
...
+const graylist = {
+  addresses: process.env.IP_GRAYLIST && process.env.IP_GRAYLIST.split(',')
+}
+
+if (graylist.addresses) {
+  graylist.authorizedAddrs = []
+  graylist.authorizedBlocks = []
+
+  graylist.addresses.forEach((entry) => {
+    if ((entry.indexOf('/') === -1) && (entry.split('.').length === 4)) return graylist.authorizedAddrs.push(entry)
```
diracdeltas: if you're trying to validate an ipv4 address here, you can just use https://www.npmjs.com/package/validate-ip-node which is already a dep of this library

evq: there's a lot of places where we can / should clean up code, but for this PR I asked Michael to focus on making the existing code work with the latest hapi with as few changes to make it easier to verify. for something like this I'd prefer we open an issue and address in a follow up PR

bat-utils/lib/hapi-server.js
```diff
@@ -41,13 +47,14 @@ function pushTokens (map) {
...
+async function Server (options, runtime) {
   const debug = new SDebug('web')
 
-  const graylist = { addresses: process.env.IP_GRAYLIST && process.env.IP_GRAYLIST.split(',') }
-  const server = new hapi.Server()
-
-  server.connection({ port: process.env.PORT })
+  const serverOpts = {
+    port: options.port,
+    host: '0.0.0.0'
```
diracdeltas: why 0.0.0.0 instead of 127.0.0.1?

ledger/controllers/grants.js
```diff
@@ -152,8 +152,9 @@ const safetynetPassthrough = (handler) => (runtime) => async (request, reply) =>
     })
   } catch (e) {
     try {
+      console.log(e.data.payload.toString())
```
diracdeltas: extra console.log?

bat-utils/lib/hapi-rate-limiter.js
maikelmclauflin: https://github.com/brave-intl/bat-ledger/issues/743

ledger/controllers/grants.js
maikelmclauflin: thanks for that

bat-utils/lib/hapi-server.js
maikelmclauflin: i believe this was to work with docker containers

```The host property set to localhost is likely the safest choice. In a docker container, however, the localhost may not be accessible outside of the container and using host: '0.0.0.0' may be needed.
```
https://hapi.dev/tutorials/

just ran the tests with this change locally, got a connection refused error

diracdeltas:
lgtm now

PR #732, opened 2019-10-07
added scoping keys to each test

maikelmclauflin:


bat-utils/lib/hapi-server.js
```diff
@@ -101,6 +103,7 @@ async function Server (options, runtime) {
 
   debug('extensions registered')
 
+  console.log(runtime.login)
```
nvonpentz: log line

bat-utils/lib/hapi-server.js
```diff
@@ -30,6 +30,8 @@ module.exports = async (options, runtime) => {
 const pushScopedTokens = pushTokens({
   TOKEN_LIST: 'global',
   // env var       // scope key
+  ALLOWED_STATS_TOKENS: 'stats',
+  ALLOWED_REFERRALS_TOKENS: 'referrals',
```
nvonpentz: As far as I can tell, besides docker and travis configs, this is the only non-test file that's edited in this PR.  It also looks like `ALLOWED_ADS_TOKENS` and `ALLOWED_PUBLISHERS_TOKEN` were already in the codebase.  So am I correct in thinking that the code change that has the server verify the scoped tokens of incoming requests happened in an earlier PR, and this just focuses on the tests?

maikelmclauflin: yeah. code changes happened a while ago. this is just to confirm in the tests that all of the routes are using their appropriate scopes instead of assuming that they are since all of the keys were previously the same

PR #759, opened 2019-11-19
threw failing v1/grants to stop exec

maikelmclauflin:


evq:
obe - https://github.com/brave-intl/bat-ledger/pull/805

PR #784, opened 2019-12-11
forwards wallet claims to grants

maikelmclauflin:
in order to spin down ledger we'll need to migrate responsibilities elsewhere. linking wallets belongs where wallets will belong, which currently is the grant server


evq:
closing as this has been moved to https://github.com/brave-intl/bat-ledger/pull/907

PR #795, opened 2020-01-28
modified virtual grant groupings

maikelmclauflin:
virtual grants are now segmented by date. this allows us to do daily settlements :)
allows us to freeze virtual grant surveyors

### DIFF
--- a/eyeshade/workers/reports.js
+++ b/eyeshade/workers/reports.js
@@ -49,10 +49,23 @@ async function freezeOldSurveyors (debug, runtime, olderThanDays) {
   `
 
   const {
-    rows
+    rows: nonVirtualSurveyors
   } = await runtime.postgres.query(query, [olderThanDays])
 
-  await Promise.all(rows.map(async (row) => {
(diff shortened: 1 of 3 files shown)


eyeshade/workers/reports.js
```diff
@@ -49,10 +49,23 @@ async function freezeOldSurveyors (debug, runtime, olderThanDays) {
   `
```
quocviet339: 2000

eyeshade/workers/reports.js
```diff
@@ -49,10 +49,23 @@ async function freezeOldSurveyors (debug, runtime, olderThanDays) {
...
 
   const {
-    rows
+    rows: nonVirtualSurveyors
   } = await runtime.postgres.query(query, [olderThanDays])
 
-  await Promise.all(rows.map(async (row) => {
+  const virtualQuery = `
```
evq: should we maybe add a test to ensure the virtual surveyor freezing is working?

PR #813, opened 2020-03-06
added tests for peer to peer referrals

maikelmclauflin:
* peer to peer referrals will not send an owner id
* remove validation, and check transaction output from finalized referrals

### DIFF
--- a/eyeshade/controllers/referrals.js
+++ b/eyeshade/controllers/referrals.js
@@ -28,7 +28,7 @@ const originalRateId = '71341fc9-aeab-4766-acf0-d91d3ffb0bfa'
 
 const amountValidator = braveJoi.string().numeric()
 const groupNameValidator = Joi.string().optional().description('the name given to the group')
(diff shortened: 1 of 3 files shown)


eyeshade/controllers/referrals.js
```diff
@@ -28,7 +28,7 @@ const originalRateId = '71341fc9-aeab-4766-acf0-d91d3ffb0bfa'
 
 const amountValidator = braveJoi.string().numeric()
 const groupNameValidator = Joi.string().optional().description('the name given to the group')
-const publisherValidator = braveJoi.string().publisher().required().description('the publisher identity')
+const publisherValidator = braveJoi.string().publisher().allow(null, '').optional().description('the publisher identity')
```
yachtcaptain23:
```suggestion
const publisherValidator = braveJoi.string().publisher().allow(null, '').optional().description('the publisher identity. e.g. youtube#VALUE, twitter#VALUE, reddit#value, etc., or null.  owner aka publishers#VALUE should not go here')
```

PR #821, opened 2020-03-12
added snapshot endpoint and worker to fill the table

maikelmclauflin:
added endpoint to help antifraud and publishers create snapshots of user balances for faster queries


eyeshade/workers/snapshots.js
```diff
@@ -0,0 +1,94 @@
+const Cursor = require('pg-cursor')
+const uuidV5 = require('uuid/v5')
+const { BigNumber } = require('bat-utils/lib/extras-utils')
+
+exports.initialize = async (debug, runtime) => {
+  await runtime.queue.create('update-snapshot-accounts')
+}
+
+const getAllAccountBalances = `
+select * from account_balances
```
nvonpentz: I think ideally we would be able to generate account balances for any period in time (including the past).  This query only supports generating snapshots at the given moment.  So I think if we bypass account_balances and go straight to account_transactions, we can do this.

This is a query I use to generate ads account balances precisely on the first one the month - https://github.com/brave-intl/gordon/blob/b8e52fb77244a1de0816bd073bb5646236f0b490/bin/create-payout-report-ads-v2.py#L16.  I think some version of that query can work here.  Then we can let publishers (or whatever client ends up creating these snapshots) determine at what time they want the balances to be frozen at.  But I wouldn't focus too much on the API for now

maikelmclauflin: yeah, started on this aspect on another revision of this. was going to run a test with prod data locally to see how slow queries would be

eyeshade/workers/snapshots.js
```diff
@@ -0,0 +1,147 @@
...
+async function createDailySnapshot (debug, runtime, date, next) {
+  const until = justDate(new Date(date))
+  const snapshotId = uuidV5(until, 'dc0befa2-37a4-4235-a5ce-dfc7d5408a78').toLowerCase()
+  const result = await updateSnapshotAccounts(debug, runtime, {
+    snapshotId,
+    until
+  })
+  next && next(debug, runtime)
+  return result
+}
```
nvonpentz: So if we have Publishers driving the generation of the snapshots each month, I don't think I can see a use case for regularly generating snapshots.  I think we can remove the cron code for now

eyeshade/workers/snapshots.js
```diff
@@ -0,0 +1,147 @@
...
+    , message      :
+      { snapshotId : 'uuid',
+        until      : '2020-01-01T00:00:00.000Z'
+      }
+    }
+  */
+  'update-snapshot-accounts': updateSnapshotAccounts
+}
+
+function setupQuery (until = 'current') {
```
nvonpentz: probably want to be consistent with naming for `until` and `latest_transaction` whatever we end up naming that

eyeshade/workers/snapshots.js
```diff
@@ -0,0 +1,147 @@
...
+    args
+  } = setupQuery(until)
+  let total = new BigNumber(0)
+  let count = new BigNumber(0)
+  const { rows } = await roClient.query(query, args)
+  roClient.release()
+  try {
+    const now = new Date()
+    await client.query('BEGIN')
+    for (let i = 0; i < rows.length; i += 1) {
```
nvonpentz: I think we can avoid this loop and use only one query for both the select and insert by doing something like
```
insert into balance_snapshot_accounts (account_id, account_type, balance)
select
  account_transactions.account_type as "accountType",
  account_transactions.account_id as "accountId",
  coalesce(sum(account_transactions.amount), 0.0) as "balance"
from account_transactions
  where created_at < $1
group by (account_transactions.account_type, account_transactions.account_id);
```
(query is incomplete)

https://www.postgresqltutorial.com/postgresql-insert/

eyeshade/migrations/0012_balance_snapshots/up.sql
```diff
@@ -0,0 +1,25 @@
...
+  created_at         timestamp with time zone not null default current_timestamp,
+  updated_at         timestamp with time zone not null default current_timestamp,
+  latest_transaction timestamp with time zone,
+  completed          boolean default false
+);
+
+create table balance_snapshot_accounts(
+  id           uuid primary key,
+  created_at   timestamp with time zone not null default current_timestamp,
+  snapshot_id  uuid not null references balance_snapshots(id),
```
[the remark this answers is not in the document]
nvonpentz: What do you think about a `payout_reports` table (what you have as `balance_snapshots`) and a `balance_snapshots` table (what you have a `balance_snapshot_accounts`) that references the payout_report_id.  So a payout report is comprised of many balance snapshots.

I'm just thinking cause `balance_snapshot_accounts` is a bit of a funny name

eyeshade/controllers/snapshots.js
```diff
@@ -0,0 +1,146 @@
...
+    if (!snapshot.completed) {
+      roClient.release()
+      return h.response(snapshot).code(202)
+    }
+    const { rows: accountBalances } = await roClient.query(getSnapshotBalanceAccounts, [snapshotId])
+    snapshot.items = accountBalances
+    roClient.release()
+    return snapshot
+  }
+}
```
nvonpentz: I think what we want from the GET is to expose the ability for publishers to get the balance of a specific account_id within a particular snapshot, rather than bulk GET endpoint to get the entire snapshot.

This basically would be used on the Publishers admin dashboard so admins could go to a particular publisher's page and see 'oh this account is supposed to receive 30 BAT this settlement'

eyeshade/workers/snapshots.js
```diff
@@ -0,0 +1,99 @@
...
+const writeAccountBalancesBefore = `
+insert into balance_snapshots (snapshot_id, account_id, account_type, balance)
+  select
+    $1 as snapshot_id,
+    account_transactions.account_id,
+    account_transactions.account_type,
+    coalesce(sum(account_transactions.amount), 0.0) as "balance"
+  from account_transactions
+    where created_at < $2
+group by (account_transactions.account_type, account_transactions.account_id);
```
nvonpentz: Indentation looks off on the `group by`

eyeshade/migrations/0012_balance_snapshots/up.sql
```diff
@@ -0,0 +1,26 @@
...
+  completed             boolean default false
+);
+
+create table balance_snapshots(
+  id           uuid primary key default uuid_generate_v4(),
+  created_at   timestamp with time zone not null default current_timestamp,
+  snapshot_id  uuid not null references payout_reports(id) on delete cascade,
+  account_id   text not null,
+  account_type text not null,
+  balance      numeric(28, 18) not null default 0.0
```
nvonpentz: I don't think this is hugely significant but I think we may not want to default to 0.0 here, since we want to be sure that whatever value is saved as balance reflects exactly what's in their account

PR #822, opened 2020-03-13
AC Votes Eyeshade

husobee:
updates to add ac vote retrieval from kafka, and insertion into the votes table


eyeshade/workers/acvote.js
```diff
@@ -0,0 +1,73 @@
+const { votesId } = require('../lib/queries.js')
+const { voteType } = require('../lib/vote.js')
+
+const voteTopic = process.env.ENV + '.payment.vote'
+
+module.exports = (runtime, callback) => {
+  runtime.kafka.on(voteTopic, async (messages) => {
+    const client = await runtime.postgres.connect()
+    const now = new Date()
+    const date = [
```
corymcdonald, line 10: The JS Date library has a lot of weirdness, especially around timezones is there any worry here? `Moment.js` might be a better option? Thoughts?

eyeshade/workers/acvote.js
```diff
@@ -0,0 +1,73 @@
...
+            publisher,
+            surveyorId
+          ])
+        }
+      } catch (e) {
+        await client.query('ROLLBACK')
+        runtime.captureException(e, { extra: { topic: voteTopic } })
+        throw e
+      }
+      await client.query('COMMIT')
```
corymcdonald, line 65: When this raises an exception due to a lock or db constraint do we capture that error somewhere?

eyeshade/workers/acvote.js
husobee: https://github.com/brave-intl/bat-ledger/pull/822/commits/8264f629ed6b80b9459222d3da822b21226ca11b <- addressed here, also updated suggestions to match.

eyeshade/workers/acvote.js
husobee: addressed in https://github.com/brave-intl/bat-ledger/pull/822/commits/8264f629ed6b80b9459222d3da822b21226ca11b logging out if there were errors from commit/rollback.

PR #844, opened 2020-04-21
resolve country groups

maikelmclauflin:
dedupes future additions to the countries / groups


eyeshade/migrations/0016_geo_referral/up.sql
```diff
@@ -0,0 +1,5 @@
+select execute($$
+
+insert into migrations (id, description) values ('0016', 'geo_referral');
+
```
evq, line 4: @maikelmclauflin I think we discussed adding a unique constraint on (created_at, country_code) just in case

maikelmclauflin: fixed in: dee2446

PR #882, opened 2020-05-22
updated query to not blow up balances

maikelmclauflin:


corymcdonald:
Are there any tests we can write that ensure that balances are valid when being returned from this endpoint?

corymcdonald:
For my own notes / others notes

Original account_balances materialized view generated here:
https://github.com/brave-intl/bat-ledger/blob/6cf687cb844a30d0e91d82ab88e318d3a65b189f/eyeshade/migrations/0003_ads/up.sql#L80-L96


Original query was this:
https://github.com/brave-intl/bat-ledger/blob/2cf204e31687a7478a4c30f54ecf93a7f494c969/eyeshade/controllers/accounts.js#L25-L29

eyeshade/controllers/accounts.js
```diff
@@ -24,26 +24,12 @@ const joiBAT = braveJoi.string().numeric()
...
-  WHERE to_account = any($1::text[])
-  GROUP BY to_account_type, to_account
-) AS q
-GROUP BY (q.account_type, q.account_id)
+  account_transactions.account_type as account_type,
+  account_transactions.account_id as account_id,
+  COALESCE(SUM(account_transactions.amount), 0.0) as balance
+FROM account_transactions
+WHERE account_id = any($1::text[])
+GROUP BY (account_transactions.account_id, account_transactions.account_type);
```
corymcdonald, line 32: I know that logically this should be the same but can we reverse this like so? Just to maintain the same structure as the first materialized view.

```suggestion
GROUP BY (account_transactions.account_type, account_transactions.account_id);
```

maikelmclauflin:
> Are there any tests we can write that ensure that balances are valid when being returned from this endpoint?

yeah, we certainly can. let me do that quick

PR #888, opened 2020-06-05
removed ro pool

maikelmclauflin:
removed readonly


maikelmclauflin:
closing because database attachment locking resolved and update is no longer needed

PR #889, opened 2020-06-08
Filter transactions

### LINKED ISSUE #883: Last Settlement endpoint
Marking as P2, as this is causing some users unable to view their page. [You can see the trace here](https://rpm.newrelic.com/accounts/2604309/applications/592980707_h592981222/transactions?type=app#id=5b22436f6e74726f6c6c65722f7075626c6973686572732f686f6d65222c22225d&app_trace_id=196a0680-9f50-11ea-a19b-0242ac110007_4188_10443&tab-detail_196a0680-9f50-11ea-a19b-0242ac110007_4188_10443=trace_details)

Basically we'd want to do this all on Eyeshade's side as a new endpoint:

```
      return {} if transactions.blank?
       # Find most recent settlement transaction
      last_transaction = transactions.select { |transaction|
        transaction["settlement_amount"].present?
      }.last # Eyeshade returns transactions ordered_by created_at

      return {} if last_transaction.nil?

      last_settlement_time = last_transaction["created_at"].to_time # Eyeshade returns transactions ordered_by created_at

       # Find all settlement transactions that occur within the same month as the last settlement timestamp
      last_settlement_transactions = transactions.select { |transaction|
        transaction["created_at"].to_time.at_beginning_of_month == last_settlement_time.at_beginning_of_month &&
        transaction["settlement_amount"].present?
      }

      # Sum the all the settlement transactions
      last_settlement_currency = last_settlement_transactions&.first["settlement_currency"] || nil

      # The amount is the BAT that was withdrawn from our ledger. This is a negative value.
      last_settlement_amount_bat = last_settlement_transactions.map { |transaction|
        transaction["amount"].to_d.abs
      }.reduce(:+)

      # Settlement amount will be in the user's default currency
      last_settlement_amount_currency = last_settlement_transactions.map { |transaction|
        transaction["settlement_amount"].to_d
      }.reduce(:+)

      {
        amount_bat: last_settlement_amount_bat,
        amount_currency: last_settlement_amount_currency,
        timestamp: last_settlement_time.to_i,
        currency: last_settlement_currency
      }
```

maikelmclauflin:
* run tests inside of docker container (odd errors occuring outside only on travis)
* transform transactions on eyeshade side
fixes #883


corymcdonald:
When I ran the following query the execution time took ~ 37 seconds.
```sql
EXPLAIN (ANALYZE, COSTS, VERBOSE, BUFFERS, FORMAT JSON)
SELECT
  created_at, description, channel, amount, settlement_currency, settlement_amount, transaction_type
FROM account_transactions
WHERE account_id = 'publishers#uuid:d56972c2-4933-48bb-a7f0-9cfcfdd47077'
AND transaction_type = ANY('{referral_settlement, contribution_settlement}'::text[])
```

image


It looks like `account_transactions` is a materialized view. I was looking into what we could do to speed this up and found this really interesting concept introduced in Postgres 11 called a [covering index](https://www.postgresql.org/docs/11/indexes-index-only-scans.html). It looks like it works by indexing some attributes but then including others on top of the existing properties. Something like:

```sql
CREATE INDEX transactions_filter_idx 
ON account_transactions (account_id,transaction_type) 
INCLUDE (created_at, description, channel, amount, settlement_currency, settlement_amount, transaction_type);
```

The one thing I'm concerned about is that since this is a materialized view when we refresh it might take a lot longer to build this index. What are your thoughts?

eyeshade/migrations/0017_include_indexes/up.sql
```diff
@@ -0,0 +1,13 @@
+select execute($$
+
+insert into migrations (id, description) values ('0017', 'include_indexes');
+
+CREATE INDEX transactions_filter_idx
+ON transactions (from_account, to_account)
+INCLUDE (created_at, description, channel, amount, settlement_currency, settlement_amount, transaction_type);
+
+CREATE INDEX transactions_with_type_filter_idx
```
corymcdonald, line 9: I think we can only create this index. Postgres should be smart enough to use ignore any additional fields. 👍

eyeshade/controllers/accounts.js
```diff
@@ -53,27 +53,79 @@ v1.getTransactions =
...
+        }
+        if (settlementAmount) {
+          transaction.settlement_amount = settlementAmount
+        }
+        if (settlementTypes[transactionType]) {
+          if (toAccountType) {
+            transaction.settlement_destination_type = toAccountType
+          }
+          if (toAccount) {
+            transaction.settlement_destination = toAccount
```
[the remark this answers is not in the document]
maikelmclauflin: basically i am just replicating exactly what was being done with the view, but in a map. so in this case only for transactions that are a type of settlement and since there is no settlement coming from an individual other than ourselves, i didn't include that condition. basically the view uses the to_account field: 
https://github.com/brave-intl/bat-ledger/blob/master/eyeshade/migrations/0001_transactions/up.sql#L47
https://github.com/brave-intl/bat-ledger/blob/master/eyeshade/migrations/0001_transactions/up.sql#L61

eyeshade/migrations/0017_include_indexes/up.sql
maikelmclauflin: fair enough. fixed in 1dcaf97

PR #896, opened 2020-06-15
handle kafka handler throws

maikelmclauflin:


maikelmclauflin:
effects became redundant

PR #898, opened 2020-06-16
Topic with incremental offset commital

maikelmclauflin:
* isolates kafka topics so that if one fails the others don’t
* ~pushes forward a single topic as far as possible so that we don’t retry a batch of 10 when only 5 complete and the 6th fails~
* flip a bool passed to the consume function to have sinek not convert to string first so that we don’t get bad number deserialization errors. (`80` -> `-127`)

### DIFF
--- a/bat-utils/lib/runtime-kafka.js
+++ b/bat-utils/lib/runtime-kafka.js
@@ -33,7 +33,7 @@ class Kafka {
 
     this.producer = new NProducer(this.config, null, partitionCount)
     this.producer.on('error', error => {
-      console.error(error)
+      console.error('error handler', error)
       if (this.runtime.captureException) {
         this.runtime.captureException(error)
       }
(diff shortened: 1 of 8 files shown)


bat-utils/lib/runtime-kafka.js
```diff
@@ -57,21 +57,39 @@ class Kafka {
...
-          debug('batch', topic, messages.length, messages[0])
-          return this.topicHandlers[topic](messages)
-        })
-
-        // wait until all partitions of this topic are processed and commit its offset
-        // make sure to keep batch sizes large enough, you dont want to commit too often
-        await Promise.all(partitionPromises)
-        await consumer.commitLocalOffsetsForTopic(topic)
+          const partLength = partition.length
+          const offset = partition[partition.length - 1].offset
```
evq, line 65: I'm worried this isn't the right offset for us to be committing. [commitLocalOffsetsForTopic appears to work quite differently](https://github.com/nodefluent/node-sinek/blob/a90ba6c1b9d007c4d9577853500e68bba201c60c/lib/librdkafka/NConsumer.js#L894) - I would think we would want essentially the same logic it uses to determine the offset but only for a single topic / partition?

evq: Michael and I caught up about this - I got mixed up, here `partition` is the batch of messages from that partition - which has embedded offsets. so I believe this logic would work

bat-utils/lib/runtime-kafka.js
```diff
@@ -57,25 +57,44 @@ class Kafka {
       // parallel processing on topic level
       const topicPromises = Object.keys(batchOfMessages).map(async (topic) => {
         // parallel processing on partition level
```
evq, line 59: This is no longer accurate, but we could switch back to processing the partitions in parallel - right? Since we're saving the offset on a (topic, partition) basis.

bat-utils/lib/runtime-kafka.js
```diff
@@ -57,25 +57,44 @@ class Kafka {
...
-
-        // wait until all partitions of this topic are processed and commit its offset
-        // make sure to keep batch sizes large enough, you dont want to commit too often
-        await Promise.all(partitionPromises)
-        await consumer.commitLocalOffsetsForTopic(topic)
+          for (let j = 0; j < partitionMessages.length; j += 1) {
+            const message = partitionMessages[j]
+            try {
+              await handler(message)
+              consumer.commitOffsetHard(topic, partitionKey, message.offset)
```
evq, line 71: does it make sense to defer the offset commit until we've either successfully processed all messages for a partition in this batch or gotten an error while processing one? that way we aren't roundtripping with kafka every time we process a single message

bat-utils/lib/runtime-kafka.js
maikelmclauflin: that is probably correct

bat-utils/lib/runtime-kafka.js
maikelmclauflin: yeah, i can refactor to do that

PR #909, opened 2020-07-02
added migration for wallets

maikelmclauflin:
* takes wallets from ledger
* upserts to grants

### DIFF
--- a/bin/migrate-wallets-to-grants.js
+++ b/bin/migrate-wallets-to-grants.js
@@ -9,6 +9,7 @@ DEBUG=-* DATABASE_URL= MONGODB_URI= ./bin/migrate-wallets-to-grants.js
 const bson = require('bson')
 const Postgres = require('bat-utils/lib/runtime-postgres')
 const { MongoClient } = require('mongodb')
+const url = require('url')
 
 const insertStatement = `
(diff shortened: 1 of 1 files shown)


bin/migrate-wallets-to-grants.js
```diff
@@ -0,0 +1,156 @@
...
+function objectIdFromTimestamp (_timestamp) {
+  // Convert string date to Date object (otherwise assume timestamp is a date)
+  const timestamp = new Date(_timestamp)
+  // Convert date object to hex seconds since Unix epoch
+  var hexSeconds = Math.floor(timestamp / 1000).toString(16)
+  // Create an ObjectId with that hex timestamp
+  return bson.ObjectID.createFromHexString(hexSeconds + '0000000000000000')
+}
+
+function pathname (url) {
```
diracdeltas, line 115: can you just construct a node URL object and get the pathname there? https://nodejs.org/api/url.html#url_url_pathname

maikelmclauflin: unfortunatley, i tried this, and the reason i went around was to because whenever urls have a shard as many mongo ones on heroku do, an `ERR_INVALID_URL` error is sent back when parsing the url. however, there is the parse member which can be used for this purpose with an eslint disable comment as done: 
bf49d4d

PR #930, opened 2020-07-29
fixed deps

maikelmclauflin:
* fixes joi dep missing (https://www.npmjs.com/package/@hapi/joi)
* fixes vulns from npm audit


maikelmclauflin:
@corymcdonald unfortunately if you look in the travis log ava fails. uses a different syntax. need to remove anonize before we can move forward

PR #945, opened 2020-10-06
added migration for inserted at and modified tests

maikelmclauflin:
- [x] adds migration for inserted_at column to help with antifraud etl'ing
- [x] update tests to account for new column

### DIFF
--- a/eyeshade/migrations/0019_add_inserted_at/up.sql
+++ b/eyeshade/migrations/0019_add_inserted_at/up.sql
@@ -1,0 +1,12 @@
+select execute($$
+
+insert into migrations (id, description) values ('0019', 'add_inserted_at');
+
+alter table transactions add column
+inserted_at timestamp with time zone
+not null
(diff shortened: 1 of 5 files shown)


eyeshade/migrations/0019_add_inserted_at/up.sql
```diff
@@ -0,0 +1,12 @@
+select execute($$
+
+insert into migrations (id, description) values ('0019', 'add_inserted_at');
+
+alter table transactions add column
+inserted_at timestamp with time zone
+not null
+default current_timestamp;
+
+update transactions set inserted_at = created_at;
```
nvonpentz, line 10: I worry about this since I think this will create a lock on every row in the transactions table until the statement completes

maikelmclauflin: unsure, but we can take it out

nvonpentz (approved):

PR #956, opened 2020-11-05
allowed partition constraint to be passed

maikelmclauflin:
in order to not block the current stream of messages going by we created a second partition to add old messages to so that they can be processed as throughput is available


husobee (approved):
Assuming the live data producer is configured to use partition 0 for writes.

PR #964, opened 2020-11-10
Disable mongo

maikelmclauflin:
disables mongo index creation to not prevent app from starting up


eyeshade/lib/settlements.js
```diff
@@ -43,7 +41,9 @@ const v2 = avro.parse({
     { name: 'fee', type: 'string' },
     { name: 'commission', type: 'string' },
     { name: 'fees', type: 'string' },
-    { name: 'type', type: 'string' }
+    { name: 'type', type: 'string' },
+    { name: 'hash', type: 'string' },
```
evq, line 45: is it okay to reorder hash like this?

maikelmclauflin: it hasn't been used yet. so yes

evq (approved):

PR #971, opened 2020-11-14
reduced number of queries

maikelmclauflin:


eyeshade/workers/referrals.js
```diff
@@ -103,24 +108,30 @@ module.exports.consumer = (runtime) => {
...
+      return {
+        id,
+        referral
+      }
+    })
+    const ids = docs.map(({ id }) => id)
+    const {
+      rows: previouslyInserted
+    } = await postgres.query(getTransactionsById, [ids])
+    return Promise.all(docs.map(async ({ id: targetId, referral }) => {
```
evq, line 120: should we be `await`ing this instead of returning?

maikelmclauflin: it ends up being the same thing because we are `await` ing on the other side and this is an `async` function

evq (approved):

PR #974, opened 2020-11-18
added owner filter

maikelmclauflin:
- add `publishers#uuid:` prefix if none exists
- disallow non null owners with id of `removed`
- use `finalizedTimestamp` for created at instead of timestamp put on kafka queue


eyeshade/workers/referrals.js
```diff
@@ -113,11 +114,16 @@ module.exports.consumer = (runtime) => {
         referral
       }
     })
-    const ids = docs.map(({ id }) => id)
+
+    // drop documents that do not start with 'publishers#uuid:' or have a 'removed' id
+    const filteredDocs = docs.filter(({ referral: { _id: { owner } } }) => {
+      return !owner || (owner.slice(0, 16) === 'publishers#uuid:' && owner.slice(16) !== 'removed')
```
keur, line 120: i think your `&&` should be an `||`

eyeshade/workers/referrals.js
```diff
@@ -117,7 +117,7 @@ module.exports.consumer = (runtime) => {
 
     // drop documents that do not start with 'publishers#uuid:' or have a 'removed' id
     const filteredDocs = docs.filter(({ referral: { _id: { owner } } }) => {
-      return !owner || (owner.slice(0, 16) === 'publishers#uuid:' && owner.slice(16) !== 'removed')
+      return !owner || owner.slice(0, 16) === 'publishers#uuid:' || owner.slice(16) !== 'removed'
```
keur, line 120: Do you think maybe that we should log a message to sentry? It seems like we should know if promo is sending badly formatted values to eyeshade

keur (approved):

evq (approved):

PR #988, opened 2020-12-16
added redis connection string building when password is set

maikelmclauflin:
if `REDIS_PASSWORD` is available as in the case of the k8s cluster, we need to dynamically build the string because we cannot do so in the yaml files in the eyeshade-ops repo

### DIFF
--- a/config.js
+++ b/config.js
@@ -39,6 +39,11 @@ if (!service) {
 }
 
 process.env.PORT = process.env.PORT  || service.portno
+let redisURL = process.env.REDIS_URL
+const redisPass = process.env.REDIS_PASSWORD
+if (redisPass) {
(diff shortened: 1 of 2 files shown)


config.js
```diff
@@ -39,6 +39,11 @@ if (!service) {
 }
 
 process.env.PORT = process.env.PORT  || service.portno
+let redisURL = process.env.REDIS_URL
+const redisPass = process.env.REDIS_PASSWORD
+if (redisPass) {
+  redisURL = `redis://user:${redisPass}@${process.env.REDIS_MASTER_SERVICE_HOST_REDIS}:${process.env.REDIS_MASTER_SERVICE_PORT_REDIS}`
```
husobee, line 45: I would suggest we url encode the password if we are building the connection string.  I would prefer if the env variable is the connection string though.

husobee (changes requested):

config.js
maikelmclauflin: i agree, i just haven't found a way around this currently. does not seem to connect when using 
```
redis://user:PASS@redis:6379
```

husobee (approved):

PR #989, opened 2020-12-16
Put back endpoints and cleaned

maikelmclauflin:
channeled my inner @evq and decided it was probably best to leave the endpoints as 410
i think deleting the code is still the right move, so i moved all of the endpoints under an array in hapi-server
beyond that, just cleanup


husobee (approved):
lgtm

PR #1003, opened 2021-03-10
updates to include a seperate worker process

husobee:


eyeshade/worker.js
```diff
@@ -0,0 +1,25 @@
...
+if (!process.env.BATUTIL_SPACES) {
+  process.env.BATUTIL_SPACES = '*,-hapi'
+}
+
+Runtime.newrelic.setupNewrelic(config, __filename)
+
+const runtime = new Runtime(config)
+
+// call the report worker initialize
+reportWorker(true, runtime)
```
evq, line 23: I think this might need to be `reportWorker.initialize` because of how we're exporting in `workers/reports`. it may be easiest if we revert parts of ddea44db9651bd843bb8fb27e233cbd1b33fcc06 to get this going again. I think we also had the worker in the docker-compose which will make it easier to test that it's working

evq (changes requested):

eyeshade/worker.js
husobee: agree. closing pr.

PR #1008, opened 2021-03-15
parallelized insertions and freezing

maikelmclauflin:
- parallelized functional insertions (1 round trip currently takes ~140ms) turned into batches of 25
- added document_id index to tx table with note about concurrency requirement

### DIFF
--- a/eyeshade/migrations/0019_add_inserted_at/down.sql
+++ b/eyeshade/migrations/0019_add_inserted_at/down.sql
@@ -3,8 +3,6 @@ select execute($$
 alter table if exists transactions
 drop column if exists inserted_at;
 
-drop index if exists transactions_inserted_at;
-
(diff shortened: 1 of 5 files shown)


eyeshade/migrations/0019_add_inserted_at/down.sql
```diff
@@ -3,8 +3,6 @@ select execute($$
 alter table if exists transactions
 drop column if exists inserted_at;
 
-drop index if exists transactions_inserted_at;
```
maikelmclauflin, line 6: noted that this left a notice in logs:
```
psql:./migrations/0019_add_inserted_at/down.sql:10: NOTICE:  index "transactions_inserted_at" does not exist, skipping
```

husobee (approved):

PR #1013, opened 2021-03-17
Batch surveyor txs

maikelmclauflin:
- refactor postgres code to treat inserts differently
- insert batches of transactions at the same time
- dynamically generate the query string and flatten arguments

currently a single surveyor with 4000 publishers, takes about 10 minutes to complete just the transaction insertion part.
the main cause of this is the round trip that the client does with each query. even making method calls concurrent, does not mitigate this as there is an internal mechanism for serially executing the queries. to get around this, the ability to add many transactions (only for voting for now) in a foolproof manner was added. this should bring the theoretical execution time down to about 24 seconds for this portion of the code (inserting transactions) since we are almost exclusively bottlenecked by the ordering of transactions going back and forth to the db.

some code snippets you can try yourself

pre allocated arrays do not compact
```
> const a = new Array(4)
> const b = new Array(4)
> [].concat.apply([], a)
[ undefined, undefined, undefined, undefined ]
> [].concat.apply([], [a,b])
[ <8 empty items> ]
```

get the end of a string
```
> 'batches'.slice('batches'.length - 2)
'es'
```

values section of query is generated correctly
```
> [['a', 'id'],['b', 'id']].map((row, rowIndex) => 
   `( ${row.map((_, argIndex) => `$${1 + argIndex + (row.length * rowIndex)}`).join(', ')} )`
).join(',\n')
'( $1, $2 ),\n( $3, $4 )'
```

to see the queries used in tests, feel free to use
```bash
DEBUG=*postgres ./node_modules/.bin/ava ./test/eyeshade/transaction.integration.test.js -m "insert from many voting transaction"
# and
DEBUG=*postgres ./node_modules/.bin/ava ./bat-utils/lib/runtime-postgres.test.js
```

### DIFF
--- a/eyeshade/lib/transaction.js
+++ b/eyeshade/lib/transaction.js
@@ -259,25 +259,6 @@ async function insertManual (runtime, client, settlementId, created, documentId,
   ], client)
 }
 
-function insertFromVotingQuery (rows) {
-  const longest = rows.reduce((memo, row) => row ? Math.max(row.length, memo) : memo, 0)
-  const filtered = rows.filter((row) => row)
-    .map((row) => {
-      const newRow = new Array(longest)
-      row.forEach((arg, index) => {
-        newRow[index] = arg
-      })
-      return newRow
-    })
-  const query = `
-insert into transactions ( id, created_at, description, transaction_type, document_id, from_account, from_account_type, to_account, to_account_type, amount, channel )
-VALUES`
-  return {
-    filtered,
-    query
-  }
-}
-
 function insertFromVotingArguments (settlementAddress, voteDoc, surveyorCreatedAt) {
   if (voteDoc.amount) {
     const amount = new BigNumber(voteDoc.amount.toString())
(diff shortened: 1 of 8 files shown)


eyeshade/lib/transaction.js
```diff
@@ -269,30 +273,57 @@ async function insertFromVoting (runtime, client, voteDoc, surveyorCreatedAt) {
...
         normalizedChannel
-      ], client)
+      ]
     }
   } else {
     throw new Error('Missing amount field')
   }
 }
 
+async function insertManyFromVoting (atOneTime, runtime, client, docs, surveyorCreatedAt) {
```
keur, line 296: nit: more commonly referred to as `batchSize`

eyeshade/lib/transaction.js
```diff
@@ -269,30 +273,57 @@ async function insertFromVoting (runtime, client, voteDoc, surveyorCreatedAt) {
...
+    )
+    const query = `
+    insert into transactions ( id, created_at, description, transaction_type, document_id, from_account, from_account_type, to_account, to_account_type, amount, channel )
+    VALUES %L`
+    await runtime.postgres.query(format(query, mapped), null, client, {
+      text: query,
+      length: mapped.length
+    })
+  }
+}
```
keur, line 313: You either need to put an assertion at the beginning of the function that `docs.length` is divisible by batch size, or you need special handling to insert what's left after the loop. Pulling that out into an closure is how [we do it in stats](https://github.com/brave/stats/blob/b8746ea289adc3dcc9cff2d9d3a01f68068c1a28/src/services/common-aggregation.service.js#L151-L178).

keur: resolved in DM.

keur (approved):

PR #1014, opened 2021-03-17
Prod Release 2021-03-17_0

maikelmclauflin:
* batched serial transaction into single query

* refactored insert prep

* removed unnecessary check

* converted to map completely

* removed workflow_dispatch

* added tests for the two untested levels

* fixed implementation and used prepinsert inside of insert

* made code for insertfromvoting dry

* utilized pg-format and removed extra tests

* updated batch size var


keur (approved):
++

PR #1017, opened 2021-03-18
exposed acvote insert function and used in surveyors test

maikelmclauflin:
- exposed acvote insert function
- used insert function in test for surveyors cron job
- made sure results are deterministic


husobee (approved):
looks good

PR #1021, opened 2021-03-18
Prod Release 2021-03-18_2

maikelmclauflin:
* removed currency codes check

* added extra currency check

* flipped bool

* simplified


evq:
++

PR #1040, opened 2021-07-01
Add referral group scope

maikelmclauflin:
publishers was missing from the referral group api scope. because this is a readonly endpoint, it is ok to add it


tsmartt (approved):

yachtcaptain23 (approved):
Not sure why this why this is broken, as a key rotation shouldn't break this. From a code perspective this is fine.

PR #1074, opened 2022-06-22
Chore/tyler wip suggestions

jlbyrne:
Contains changes related to upgrading dependencies and replacing sinek with kafkajs.


tsmartt (approved):

tsmartt:
Going to merge and test this out in staging.

evq:
@tsmartt did we get a security review for this? good to get check esp for pathological failure cases to ensure no accounting issues

Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM node:11.15-stretch
+FROM node:16.15-stretch
```
fmarier, line 1: This image is no longer supported (it's up to `16.16` already and `stretch` is too old a version of Debian), could we use `node:16-bullseye` instead? That way we don't have to update the `Dockerfile` every time a new point release comes out.

Alternatively `node:16.16-bullseye`, `node:16-buster`, or `node:16.16-buster` would work fine too.

jlbyrne: Yes!  See #1077

fmarier: Great, thanks!

PR #1080, opened 2022-08-01
brought node version in line with available docker version

jlbyrne:


Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM node:16-bullseye
```
tsmartt, line 1: I think we want to leave this unpinned so we get automatic updates.

tsmartt (approved):

PR #1090, opened 2022-10-24
Potential changes

tsmartt:


tsmartt:
@evq @jlbyrne I extracted some components for easier testing and added tests to core worker code.

@thypon for a sec review on the new dependency (sinonJS) for mocks/stubs.

eyeshade/lib/publishers.test.js
```diff
@@ -0,0 +1,24 @@
...
+const { hasValidCountry } = require('./publishers.js')
+const sinon = require('sinon')
+const { Runtime } = require('bat-utils')
+const config = require('../../config')
+const tRuntime = new Runtime(config)
+
+test('should return true if Exception', async (t) => {
+  const stubPost = sinon.stub().throws()
+  const val = await hasValidCountry(tRuntime, 'abc', stubPost)
+  t.deepEqual(val, true, 'Should return true if trying to make http calls blows up')
```
jlbyrne, line 11: 😆

jlbyrne (approved):
LGTM

jlbyrne (dismissed):

eyeshade/workers/suggestions.js
```diff
@@ -20,43 +20,53 @@ module.exports = (runtime) => {
...
+      await handleMessage(runtime, suggestion, client)
     }
   })
 }
+
+async function handleMessage (runtime, suggestion, client, hasValidCountryFunc = hasValidCountry) {
+  const publisher = suggestion.channel
+  const date = moment().format('YYYY-MM-DD')
+
+  if (await hasValidCountryFunc(runtime, publisher)) {
```
husobee, line 32: so, for every message we get on the consumer we will be calling the publishers api to validate if the channel belongs to an allowable country?  Will this not back up the consumers and increase the load a lot to publishers api?  Curious about the impact of doing this and what we will need to scale up.

husobee (commented):
mainly just wondering about what happens when publishers is down.  curious if we can do this check further down the line, possibly in the payout process instead (as we know the channel at payout time)

eyeshade/workers/suggestions.js
tsmartt: Yes it will increase the load on publishers, and we can work on a more elegant, cached list that runs nightly as a v2. If the consumers slow down too much, we can partition. 

We could also do it during payout, but that doesn't quite meet the requirements. 

If publishers is down, for any Exception, we let the vote through by default.

eyeshade/workers/acvote.js
```diff
@@ -18,40 +20,44 @@ module.exports = (runtime) => {
...
-  const surveyorUpdate = `
+    const surveyorUpdate = `
       insert into surveyor_groups (id, price, virtual) values ($1, $2, true)
       on conflict (id) do nothing;
       `
-  await runtime.postgres.query(surveyorUpdate, [
-    surveyorId,
-    voteValue
-  ], client)
+    await runtime.postgres.query(surveyorUpdate, [
```
thypon, line 44: I wonder which postgres library it's used here

thypon (changes requested):

eyeshade/workers/acvote.js
tsmartt: `pg`

thypon (approved):

husobee (approved):
spoke with tyler about why my prior comments are not applicable.

package.json
```diff
@@ -110,6 +110,7 @@
     "coveralls": "^3.1.1",
     "nodemon": "^2.0.19",
     "parse-prometheus-text-format": "^1.1.1",
+    "sinon": "^14.0.1",
```
diracdeltas, line 113: prefer to pin exact versions to avoid malicious updates

PR #1103, opened 2023-01-23
ES6 Module pre-updates

tsmartt:
In anticipation of moving to ES6 modules, I've updated the libs as far as they can go before the switch. I've pulled redis out since we don't actually use it that much, and the v3-> v4 lib move would require a bit of refactoring.

Resolves https://github.com/brave-intl/eyeshade-ops/issues/129


porteron (approved):
Looks good!

tsmartt:
Linked to https://github.com/brave-intl/eyeshade-ops/issues/129

PR #1122, opened 2023-03-30
remove remaining require statements

jlbyrne:


thypon:
@jlbyrne NIT: is there any way we can booth `config.js` and use a `json` or `yaml` file?

jlbyrne:
> 

Given the way config.js is set up, that would be a non-trivial change.  We could add a ticket for that in the backlog, possibly.

tsmartt (dismissed):

porteron (approved):

PR #1127, opened 2023-04-25
upgrade ratios endpoint to v2, get rid of unnecessary code

jlbyrne:


nvonpentz (approved):
Please test on staging before deploying

PR #1133, opened 2023-07-10
Fix/fast xml update

tsmartt:
Closes https://github.com/brave-intl/creators-private-issues/issues/182

The previous fastxml update resulted in a prometheus error on deploy. This adds a test to catch that earlier.


mattwilliams85 (approved):
LGTM!

PR #1179, opened 2023-09-29
renovate.json: add org-level base

thypon:


bcaller (approved):
Sets base config to new org config from https://github.com/brave-intl/renovate-config/blob/main/default.json

mihaiplesa (approved):

